PR #119, opened 2015-04-07
A bit of documentation and formatting in src/madness/world

mgr522:
Lots of changes to the documentation in src/madness/world. (More changes to come later.)

In an effort to make the code more readable, I renamed some files and split up some larger files into smaller ones. NO CODE WAS CHANGED, JUST MOVED AROUND. It compiles fine on my machine...

Regardless, can Robert, Justus, or someone else check to make sure that I didn't screw anything up before these changes are merged in? Thanks!


justusc:
Thanks for all the work. 

1. I like some of the new file names (e.g. worldfut.h -> future.h). 

2. I would like to have file names that match the class name in the file, for example MadnessException should be in madness_exception.h.

3. Also, parallel_runtime.h does not really fit with the naming scheme in other directories. The reason for having worldfwd.h is to avoid dependencies and allow us to use world.h as the library interface. Perhaps we could change the main library header files to match the link library names (e.g. MADworld.h, MADmra.h, etc.).

4. Please do not add function parameter names for parameters that are unused. I removed these to silence warnings. If you want the name there for reference, consider using c-sytle comments.

Otherwise it looks great.

mgr522:
I will take care of 2 and 3 tomorrow. What compiler are you using for #4? clang on my mac didn't emit any warnings with these changes; I'll strip them out.

justusc:
The compiler warnings pop up with higher warning levels.

mgr522:
What warning level are you using? I had -Wall on...

justusc:
-pedantic

justusc:
I do not want to say definitely change the interface headers to MADworld.h, etc. I think we should hear from others on that point before you make further changes to those file names. I do not think it should be that big an issue since we have the main madness.h header now.

mgr522:
Sounds good to me. When we have a consensus, I'll make the changes.

PR #121, opened 2015-04-12
Moved external libraries to a common directory

mgr522:
After a discussion with Robert, I moved the elemental, gtest, muParser, and tinyxml folders to a new "external_packages" folder in src. The packages were originally subdirectories of src or src/madness. Most updates here are essentially renaming files and propagating these changes throughout the package.

There are no compilation issues with the muParser, tinyxml, or gtest packages on my machine. I was unable to get elemental to compile/link before the changes; needless to say, it doesn't work after the move...

Can people please look at this version and let me know if you have any compile/link/run errors? I did not attempt to run any of the codes.

Thanks for your help.


justusc:
Did you consider where the header files of the external packages will be installed? It looks like the headers will be installed in <prefix>/include/external_packages/<package>, according to the include directives. I do not think this is the correct install layout.

Typically, I would install external package headers in something like <prefix>/include/<package>. This would leave your include directives unchanged in source. Then all that would need to change are the build include path (i.e. add -I<build>/src/external_packages/).

justusc:
"This would leave your include directives unchanged in source."

I take it back, you would still need to change includes directives with my proposed change.

mgr522:
I had not taken that into account and I think we should.

With regard to the changes I introduced, it's a bit moot. Some of the old locations were in madness, so there were lots of "#include <madness/muParser/...>", which isn't that different from the "#include <external_packages/muParser/...>".

The one exception was gtest. I didn't dig into the configuration too much, but this is not an issue. The includes are simply "<gtest/gtest>"; this was true before and after the move. Perhaps I/we can dig into gtest's configure a little more and try to duplicate that behavior.

Thoughts?

mgr522:
And git hub redacted the include lines. The first one was to "madness/muParser..."; the second was to "external_packages/muParser/...", both in < > angle brackets.

justusc:
Its a simple issue to fix. Basically we need to decide where we want to install the header files for these libraries. That will determine what we need to do everywhere else.

Does anyone else have strong feelings on the subject?

mgr522:
External package directory is now madness/external.

PR #128, opened 2015-05-22
Use C++11 features

mgr522:
Deleted some elements of boost-esque code in world, replacing them with C++11 features. This includes
- nullptr instead of 0 or NULL (not exhaustive)
- using "= default" or "= delete" on class functions
- Replace madness::enable_if_c with std::enable_if. Other variants, e.g. madness::disable_if_c, were changed to conform to the same semantics as std::enable_if.
- madness::ScopedPtr was removed, and nothing was using it.
- madness::ScopedArray<T> was replaced by std::unique_ptr<T[]>
- unused code in world/boost_checked_delete_bits.h was deleted.

Someone (perhaps Justus -- this was his suggestion) may want to investigate using default_delete instead of the code in boost_checked_delete_bits.h. This goes a little beyond my knowledge of default_delete and I don't want to royally break anything. (Or, Justus, perhaps we can discuss this further).

After the travis tests pass (or fail -> corrections -> pass), I will merge the changes into the master branch.


src/madness/mra/operator.h
```diff
@@ -1058,7 +1057,7 @@ namespace madness {
         /// @param[in]	key	hi-dim key
         /// @return		a lo-dim part of key; typically first or second half
         template<size_t FDIM>
-        typename disable_if_c<FDIM==NDIM, Key<NDIM> >::type
+        typename disable_if<FDIM==NDIM, Key<NDIM> >::type
```
justusc: This looks incorrect.

mgr522:
std::enable_if is the same as (what was) madness::enable_if_c. madness:enable_if was a workaround so that you didn't have to type "::value" at the end of, say, is_archive<TYPE>.

To keep things consistent, I made madness::disable_if (and the lazy variants) adhere to the same style. I added the "::value" tags where necessary.

justusc:
I think it would be better to just use std::enable_if throughout the code for consistency.

mgr522:
I was going to do that until I got to the lazy variants. It wasn't immediately clear to me how to generalize them.

Any thoughts?

PR #129, opened 2015-05-24
Checked delete upgraded to std::default_delete

mgr522:
Checked_delete (and other variants) in boost_checked_delete_bits.h have been removed; replaced by std::default_delete or lambda functions. Only a handful of places in the code actually used these functions.

Justus, please confirm that this is correct so as not to introduce memory leaks.


justusc:
Besides the minor point that we could probably use free as the deleter function without wrapping it in a lambda function, it looks good.

mgr522:
Good point. Fixed.

PR #156, opened 2015-09-01
Add CMake Support

justusc:


evaleev:
unfortunately we need yet another ci set of builds

justusc:
Yes, that is on my list of things to do.

Issue #134, opened 2015-05-25
MRA tests exceed Travis-CI time limit


justusc:
Currently all enabled tests in the MRA directory seem to run correctly, but they exceed the 50 minute time limit for Travis-CI. We may be able to improved the runtime by switching over to Travis-CI docker solution since we get more memory (4 GB, up from 3GB) and more cores (2, up from 1.5). But we need to wait for some of our package dependencies to be white listed before we can implement this solution (see issues travis-ci/travis-ci#3846, travis-ci/travis-ci#3954, travis-ci/travis-ci#3955, travis-ci/travis-ci#3956, and travis-ci/travis-ci#3957). Even with the improved performance of the new test platform, I am not optimistic that this will be enough.

naromero77:
What is Travis-CI docker solution?


On Mon, May 25, 2015 at 8:30 AM, Justus Calvin <notifications@github.com>
wrote:

> Currently all enabled tests in the MRA directory seem to run correctly,
> but they exceed the 50 minute time limit for Travis-CI. We may be able to
> improved the runtime by switching over to Travis-CI docker solution since
> we get more memory (4 GB, up from 3GB) and more cores (2, up from 1.5). But
> we need to wait for some of our package dependencies to be white listed
> before we can implement this solution (see issues travis-ci/travis-ci#3846
> <https://github.com/travis-ci/travis-ci/issues/3846>,
> travis-ci/travis-ci#3954
> <https://github.com/travis-ci/travis-ci/issues/3954>,
> travis-ci/travis-ci#3955
> <https://github.com/travis-ci/travis-ci/issues/3955>,
> travis-ci/travis-ci#3956
> <https://github.com/travis-ci/travis-ci/issues/3956>, and
> travis-ci/travis-ci#3957
> <https://github.com/travis-ci/travis-ci/issues/3957>). Even with the
> improved performance of the new test platform, I am not optimistic that
> this will be enough.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/m-a-d-n-e-s-s/madness/issues/134>.
>



-- 
Nichols A. Romero, Ph.D.

justusc:
It is a thin VM.

http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

PR #171, opened 2015-12-30
allow RUN_TEST=mra to fail since it can time out

jeffhammond:
This is my proposed solution to the problem of `mra` generating false-positives for bugs due to timing out in Travis most of the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ matrix:
       env: GCC_VERSION=4.9 RUN_TEST=tensor
     - compiler: clang
       env: GCC_VERSION=4.9 RUN_TEST=mra
+  allow_failures:
+    - env: RUN_TEST=mra
 #notifications:
 #  email:
 #    recipients:


evaleev:
it seems that timing-out is not considered a failure by travis ...

justusc:
Even with the change it looks like failures still occur. I do not want to hard code that they can fail since they will be ignored. Though the failures are being ignored anyway.

At this point, the test finish about 50% of the time. I think the better solution is to switch over to the Travis-CI docker solution. It give you 2 cores instead of 1.5. So the extra computational resources to allow the tests to finish most of the time.

justusc:
See issue #134

jeffhammond:
I've got the PRK working with Travis containers. There's a script there to
build MPI from source for Linux. It's pretty quick especially if we turn
off Romio and Fortran support.

Mac containers are better because Homebrew can do package management
without sudo.

My github activity feed will lead interested parties to the details of the
PRK implementation. Container stuff is still on a branch until I finish
cleaning up the merge.

On Wednesday, December 30, 2015, Justus Calvin <notifications@github.com>
wrote:

> See issue #134 <https://github.com/m-a-d-n-e-s-s/madness/issues/134>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/m-a-d-n-e-s-s/madness/pull/171#issuecomment-168040395>
> .
>


-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

justusc:
I created a branch with some change to allow travis containers. Current issues:
* MPICH is not allowed
* cmake is not being installed properly (can be removed for now if this cannot be resolved)
* Install the correct compiler based on the environment variable

https://github.com/m-a-d-n-e-s-s/madness/tree/travis-ci-container-build

justusc:
Last we checked, you cannot do linux and mac builds. It had to be one or the other.

jeffhammond:
Please see jeffhammond/PRK branch travis-ci-grappa. Most of the limitations
you've described here are either not true or easily solved.  I am using
MPICH 3.2b3 built from source, CMake 3.x installed from binary tar release, and
testing Mac+Linux.

On Wednesday, December 30, 2015, Justus Calvin <notifications@github.com>
wrote:

> Last we checked, you cannot do linux and mac builds. It had to be one or
> the other.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/m-a-d-n-e-s-s/madness/pull/171#issuecomment-168061494>
> .
>


-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

jeffhammond:
I fixed at least some of the issues with Travis containers.  See e.g. https://travis-ci.org/m-a-d-n-e-s-s/madness/jobs/99683427.

justusc:
I am going to close this pull request (and delete the branch) since this issue has been resolved in the other branch.

justusc:
I finally realized that we can speed up the unit tests by turning on optimizations. So with that change and moving to the container build environment MRA tests now completes in 20-30 minutes, even with the MPICH build. What a PITA this has been.

OS X builds look to be failing. I am thinking of disabling the unit tests on OS X until such time as someone wants to go in and fix them. That way we start out with a fully passing build.

jeffhammond:
I am working on the Travis stuff right now, and getting Mac+Clang to work is a goal of mine.  I will commit the appropriate `allow_failures:` so that Mac failures don't count against us, but are still tested.  Is this an acceptable compromise?  If not, can you create a branch for Linux-only testing and merge that?  I will get the Mac stuff worked out soon enough...

justusc:
As you know I have an interest in at least world working on OS X. So I would like to keep those enabled. Otherwise, allowing failures for the tensor/linalg and MRA unit tests on OS X should be fine for now. It also tests the CMake build system.

PR #182, opened 2016-02-19
mtxmq with AVX

tanutarou:
I added dynamic changing algorithm.
This program change the algorithm1 and algorithm2 by input matrix's dimensions each block.

>  A^T * B = (B^T * A)^T
>  Algorithm1: A^T * B
>  Algorithm2: (B^T * A)^T

This program selects better algorithm(better performance) of two.

And, I added (Real) * (Complex) and (Complex) * (Real) matrix multiplications with AVX.

### DIFF
--- a/src/madness/tensor/mtxmq_avx_rjh.cc
+++ b/src/madness/tensor/mtxmq_avx_rjh.cc
@@ -35,6 +35,9 @@
 #include <madness/tensor/mtxmq.h>
 #include <madness/world/worldprofile.h>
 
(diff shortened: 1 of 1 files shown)


jeffhammond:
Someone should enable this code to be tested with Travis.  I don't even know if it is integrated with the build system.

wsttiger:
This file is replacing the AVX implementation that Robert wrote. Currently it's not part of the build path. We're going to integrate this code into the mtxmq.cc.

PR #204, opened 2016-08-25
Make program compile without MPI

susilehtola:
I wrote Robert earlier for instructions on how to compile without MPI, but to no success.

Please find below a patch that adds in missing MPI stubs and CMake compile flags, after which HEAD compiles without MPI.

(The CMakeLists.txt diff is long as I removed the trailing whitespaces...)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required (VERSION 3.0.0)
 project(MADNESS CXX C ASM)
 
 # Check CMake version
-if(CMAKE_VERSION VERSION_LESS 3.5.1)  
+if(CMAKE_VERSION VERSION_LESS 3.5.1)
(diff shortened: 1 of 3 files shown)


justusc:
This functionality is already present. To compile without MPI with CMake use `-DENABLE_MPI=OFF`. The appropriate config variable is located here https://github.com/m-a-d-n-e-s-s/madness/blob/20f2f61e65af3cb0b0911441f4fee7b62cb9cc97/external/mpi.cmake#L27. Please try the above option. If it does not work, please supply the output so we can find the source of the error.

robertjharrison:
Ah ... cmake.   Sorry ... sent autoconf option.

Thanks for the patch

susilehtola:
@justusc thanks for the clarification. I undid the changes to CMakeLists.txt.

The other patches are still necessary to get MADNESS to compile with -DENABLE_MPI=OFF.

PR #221, opened 2016-12-15
added Intracomm::Split + minor improvements

evaleev:


jeffhammond:
Challenge accepted.

Should be able to review tomorrow afternoon.

justusc:
Looks good to me.

src/madness/world/safempi.h
```diff
@@ -596,6 +599,33 @@ namespace SafeMPI {
             return Intracomm(std::shared_ptr<Impl>(new Impl(group_comm, me, nproc, true)));
         }
 
+        static const int UNDEFINED_COLOR = MPI_UNDEFINED;
+        /**
+         * This collective operation creates a new \c Intracomm using
+         * the MPI_Comm_split. Must be called by all processes that
+         * belong to this communicator. Each caller must provide Color of the new Intracomm
```
evaleev: what you describe is how we currently partition communicators. the purpose of adding this was to simplify the process by replacing 2 steps with 1. it also is more general than create_group + create_comm since it allows to specify keys, but I don't see why that would matter.

PR #238, opened 2017-04-06
these files should not be in Git

jeffhammond:
@robertjharrison It appears you committed these last June.


robertjharrison:
Makes sense and Jakob approves

PR #243, opened 2017-04-13
travis: added gcc 6 to the matrix, perhaps it should replace gcc 5

evaleev:


jeffhammond:
Can we move all MRA tests to XFAIL, not just the Mac one?  Then Travis will approve.

evaleev:
@robertjharrison it's the old problem ... suddent increase in error in 1d wavepacket propagation: https://travis-ci.org/m-a-d-n-e-s-s/madness/jobs/221678854#L7601

the error is essentially stochastic so Travis is useless/helpless.

this needs to be fixed. xfailing MRA tests = giving up on large chunks of MADNESS.

robertjharrison:
I will go and revisit this test again.  I have never understood the origin
of its sensitivity.

Dunno what xfail means but all other tests are robust and failure means
either screw up or design change.

jeffhammond:
@robertjharrison XFAIL means "expected to fail".  It is standard terminology in the context of Autotools `make check`.  One usually writes at least one XFAIL test to determine that the testing infrastructure itself is working, i.e. capable of detecting failed tests.

PR #244, opened 2017-04-13
do not print config.log if there were no errors

jeffhammond:
it is bloody impossible to find the actual error in Travis logs,
particularly the raw output, when config.log is present, because it
contains a large number of innocuous instances of the word "error"
associated with configure tests that always fail.

question for reviewers: do people want to have `config.log` even when the Travis failures occur in later steps?  maybe some people do.


jeffhammond:
Yeah, I've been seeing that error for reasons I do not understand.  That shell command definitely works locally.  I'll debug.

PR #245, opened 2017-04-14
conversion of taskq::add to accept callables

evaleev:
e.g. this should work with polymorphic/generic lambdas, etc. when the interface is switched to universal refs.


evaleev:
@robertjharrison please do not merge ... due to plugging the const leak this breaks many things in TA on one of the new feature branches. currently testing a solution (variadic add with universal refs)

PR #246, opened 2017-04-19
fixed a bug for MAC OS >= 10.12

pchong90:


evaleev:
forced merge before travis checks completed since they do not use OS X 10.12

PR #249, opened 2017-05-03
prevent newbies from hard to recognize ASLR headache 

evaleev:
do runtime check in RMI::begin()

### DIFF
--- a/src/madness/world/safempi.h
+++ b/src/madness/world/safempi.h
@@ -882,6 +882,19 @@ namespace SafeMPI {
         MPI_Buffer_detach(&buffer, &size);
         return size;
     }
+
+    /// Analogous to MPI_Op_create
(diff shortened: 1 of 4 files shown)


jeffhammond:
Because user-defined datatypes are not optimized and are more likely to be broken than built-in operators, you might want to use something like the following in the future (when determining if all values are the same).
```c++
int is_equal(MPI_Comm, <type> value)
{
  <type> i[2] = { value, -value };
  <type> o[2] = { 0, 0 };
  MPI_Allreduce(i, o, 2, MPI_<TYPE>, MPI_MIN, MPI_COMM_WORLD);
  // for float types, should use fabs+threshold rather than equality
  return ( o[0] == -o[1] ) ? 1 : 0;
}
```
There is no circumstance where doubling the bytes for a scalar built-in datatype will decrease performance, unless the packet size is less than 16B.

PR #259, opened 2017-08-10
serialization of vector<T> handles non-plain T (e.g. vector<vector<T>>)

evaleev:


evaleev:
Actually might be better to split in wrap rather than in vector
serialization. That would fix array also.

PR #267, opened 2017-12-02
bare ptrs not default serializable

robertjharrison:
make bare pointers not default serializable so that pts to world object go through correct path

### DIFF
--- a/src/madness/world/type_traits.h
+++ b/src/madness/world/type_traits.h
@@ -61,7 +61,7 @@ namespace madness {
     /// This defines stuff that is serialiable by bitwise copy N.B. This reports true
     /// for \c T that is an aggregate type (struct or array) that includes pointers.
     template <typename T>
-    struct is_serializable {
+    struct is_trivially_serializable {
       static const bool value = \
         std::is_arithmetic<T>::value || \
         std::is_member_function_pointer<T>::value || \
--- a/src/madness/world/type_traits.h
+++ b/src/madness/world/type_traits.h
@@ -33,15 +33,40 @@
 #define MADNESS_WORLD_TYPE_TRAITS_H__INCLUDED
 
 #include <madness/madness_config.h>
-#  include <type_traits>
+#include <type_traits>
 
 /// \file typestuff.h
-/// \brief Grossly simplified Boost-like type traits and templates
+/// \brief type traits and templates
 
 #include <cstddef>
-#include <stdint.h>
+#include <cstdint>
 #include <madness/madness_config.h>
 #include <madness/world/function_traits.h>
+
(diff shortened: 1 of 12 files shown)


robertjharrison:
POD includes something like

struct A {
       something *ptr;
}

and nested variants thereof.   So the check for bare pointer is a functional but incomplete solution.   Overall I am happy with it.  The alternative seems to be eliminating POD entirely.

src/madness/world/type_traits.h
```diff
@@ -61,7 +61,12 @@ namespace madness {
     /// This defines stuff that is serialiable by default rules ... basically anything contiguous
     template <typename T>
     struct is_serializable {
-        static const bool value = std::is_fundamental<T>::value || std::is_member_function_pointer<T>::value || std::is_function<T>::value  || std::is_function<typename std::remove_pointer<T>::type>::value || std::is_pod<T>::value;
+        static const bool value = \
+            std::is_fundamental<T>::value || \
+            std::is_member_function_pointer<T>::value || \
+            std::is_function<T>::value  || \
+            std::is_function<typename std::remove_pointer<T>::type>::value || \
+            (std::is_pod<T>::value && !std::is_pointer<T>::value);
```
evaleev: - per http://en.cppreference.com/w/cpp/types/is_pod std::is_pod will be deprecated
- I think this clause is to make trivially-copyable composite types built of serializable elements to be serializable. There is no way right now (without reflection) to properly check this. My suggestion is to say here `(std::is_class<T>::value || std::is_array<T>::value) && std::is_trivially_copyable<T>::value` to make intent as explicit as possible.

The shortcomings of `std::is_trivially_copyable` are discussed https://groups.google.com/a/isocpp.org/forum/#!topic/std-discussion/wphImiqfX7Y%5B1-25%5D. It's a loaded topic.

robertjharrison: OK.  U R suggesting the following ??

+        static const bool value = \
+            std::is_fundamental<T>::value || \
+            std::is_member_function_pointer<T>::value || \
+            std::is_function<T>::value  || \
+            std::is_function<typename std::remove_pointer<T>::type>::value || \
+            ((std::is_class<T>::value || std::is_array<T>::value) && std::is_trivially_copyable<T>::value);

evaleev:
```
static const bool value = \
std::is_arithmetic<T>::value || \
std::is_member_function_pointer<T>::value || \
std::is_function<T>::value  || \
std::is_function<typename std::remove_pointer<T>::type>::value || \
((std::is_class<T>::value || std::is_array<T>::value) && std::is_trivially_copyable<T>::value);
```

PR #268, opened 2018-02-01
Enabled Boys localization in nemo code and fixed some inconveniences in Nemo/AC and tiny.cc

kottmanj:


fbischoff:
The tests have failed on OSX for quite a while, why? The mails I get say the tests work fine. Will approve this pull request since no additional tests fail.

PR #274, opened 2018-03-01
Futureptr as task args

evaleev:
Useful scenario: I want to use local futures (that refer to either for purely local data or refer to remote data but are cached locally), I manage their lifetime, and want to avoid unnecessary copying

### DIFF
--- a/src/madness/world/future.h
+++ b/src/madness/world/future.h
@@ -176,7 +176,7 @@ namespace madness {
     };
     template <typename T>
     struct future_to_ref<const Future<T>&> {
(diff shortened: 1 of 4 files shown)


evaleev:
This is really just an optimization for the case where Future lifetime is managed by the user. Bare ptrs are always unsafe, so Future<value*> and Future<value>* are equally unsafe. Future<std::unique_ptr<value>> might work. But this still requires std::shared_ptr<> copy ctor (and associated synchronization ... not sure how that's implemented).

PR #279, opened 2018-04-18
Polar decomposition

bsundahl:
We changed the removal of arbitrary rotations in the get_fock_transformation() routine by replacing Robert's matrix exponentiation scheme with a direct method based on the polar decomposition (See wiki link below.) 

https://en.wikipedia.org/wiki/Polar_decomposition


robertjharrison:
Breaking build is just clang/travis weirdness.

PR #291, opened 2018-08-14
Mxm

bsundahl:
Faster matrix_inner function: Makes local copies of distributed data and then computes.


fbischoff:
It compiles and runs now

PR #304, opened 2018-12-25
add madness_check and apply it where neccessary

robertjharrison:
Added MADNESS_CHECK macro that will always evaluate its argument so that MADNESS_ASSERT can be consistent with C assert which with optimization will simply vanish.  

In the test codes I changed all uses of MADNESS_ASSERT to MADNESS_CHECK.

I grepped thru all other source files and changed a handful of other use cases that either seemed to have side effects in the argument or seemed to be testing data read from a user input file

Once the pr-clean-travis has been merged into master, I will merge master into this PR so that it can be fully tested.  I have already tested it with debug and release, with 1 and 2 MPI processes.


src/madness/world/worldrmi.cc
```diff
@@ -339,9 +339,9 @@ namespace madness {
         // the worst case is where only one node sends huge messages to every node in the communicator
         // AND it has enough threads to use up all tags
         // NB list::size() is O(1) in c++11, but O(N) in older libstdc++
-        MADNESS_ASSERT(ThreadPool::size() < RMI::RmiTask::unique_tag_period() ||
-                       RMI::task_ptr->hugeq.size() <
-                       std::size_t(RMI::RmiTask::unique_tag_period() / RMI::task_ptr->comm.Get_size()));
+        MADNESS_CHECK(ThreadPool::size() < RMI::RmiTask::unique_tag_period() ||
```
evaleev: ditto: ASSERT or throw exception explicitly (or via macro)

src/madness/world/madness_exception.h
```diff
@@ -161,6 +160,33 @@ namespace madness {
...
+    do {                                        \
+        if (!(condition)) {                                             \
+            madness::exception_break(MADNESS_DISPLAY_EXCEPTION_BREAK_MESSAGE); \
+            throw madness::MadnessException("MADNESS CHECKION FAILED: " MADNESS_EXCEPTION_AT( __FILE__, __LINE__ ), \
+                                            #condition,0,__LINE__,__FUNCTION__,__FILE__); \
+        }                                                               \
+    } while (0)
+#else
+#  define MADNESS_CHECK(condition)                      \
+    do {if (!(condition)) { std::abort(); }} while (0)
```
evaleev: probably need to replace `std::abort` to `MPI_Abort` here and everywhere

src/madness/world/madness_exception.h
```diff
@@ -140,8 +140,7 @@ namespace madness {
 #endif
 
 #ifdef MADNESS_ASSERTIONS_DISABLE
-// this avoid unused variqble warnings, see http://cnicholson.net/2009/02/stupid-c-tricks-adventures-in-assert/
-#  define MADNESS_ASSERT(condition) do { (void)sizeof(condition);} while (0)
+#  define MADNESS_ASSERT(condition)
```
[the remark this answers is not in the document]
robertjharrison: Will do, but I got rid of it to be sure that condition was not evaluated --- the trick relied upon compiler optimization to detect that condition produced no side effects in order to actually eliminate it.

src/madness/world/madness_exception.h
```diff
@@ -161,6 +160,33 @@ namespace madness {
...
+/// \param[in] condition The condition to be checked.
+#define MADNESS_CHECK(condition)
+#undef MADNESS_CHECK
+
+#ifdef MADNESS_ASSERTIONS_THROW
+#  define MADNESS_CHECK(condition)              \
+    do {                                        \
+        if (!(condition)) {                                             \
+            madness::exception_break(MADNESS_DISPLAY_EXCEPTION_BREAK_MESSAGE); \
+            throw madness::MadnessException("MADNESS CHECKION FAILED: " MADNESS_EXCEPTION_AT( __FILE__, __LINE__ ), \
```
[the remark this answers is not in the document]
robertjharrison: Ugh ... sloppy.

src/madness/world/worldref.h
```diff
@@ -282,8 +282,8 @@ namespace madness {
             /// \param key The key of the \c RemoteReference object to be unregistered.
             /// \throw MadnessException If \c key is not found in the pointer map.
             static void unregister_ptr_(void* key) {
-                std::size_t ereased = pimpl_map_.erase(key);
-                MADNESS_ASSERT(ereased > 0);
+                std::size_t erased = pimpl_map_.erase(key);
+                MADNESS_CHECK(erased > 0);
```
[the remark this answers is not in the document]
robertjharrison: I agree that throw is the right thing to do.

src/madness/world/worldrmi.cc
robertjharrison: I think check would be fine here since it will always stop execution one way or another.  However, I will change to exception.

src/madness/world/madness_exception.h
robertjharrison: The motivation for std::abort was to force production of a core file.  MPI_ABORT will not routinely do this (e.g., comment from Jeff Squyres who should know https://users.open-mpi.narkive.com/wxLqT6ZJ/ompi-users-abort).   So I am inclined to leave this as is.

In re-reading the code I also noticed that if assertions were configured to do anything other than throw, that checks aborted which is I think not desirable.  So I changed the ifdef logic to make checks always throw unless assertions were configure to abort, in which case both abort.

robertjharrison:
OK, ready for next round.  Tested on my laptop with multiple MPI processes and with all possible options for assertion.

PR #307, opened 2019-01-08
Adding explicit input occupancies and stabilizing virtual states in moldft

lratcliff:
- Added the option of explicitly specifying the input occupancies in moldft, enabling core hole calculations.

- Changed the scheme for gradually decreasing the number of requested virtual states from reducing by a factor to a fixed step.  Increases flexibility and stability.


robertjharrison:
Cool! I was wanting exactly that.  I will be able to look at the PR this
weekend ... I just arrived in Japan and am buried with proposal stuff.

Also note I fixed a bug in the orthog related to virtuals so they should
now converge just fine assuming u r using canonical solver which is forced
if nva or b are nonzero.

robertjharrison:
@hborchert @lratcliff  Hannes --- can U please resurrect these changes and merge with what you did to get the virtuals working again?  This is useful functionality that I clearly let languish.

robertjharrison:
@robertjharrison test

robertjharrison:
Included in 786

PR #308, opened 2019-02-08
Mad der

bsundahl:
Adding in five new derivative operators:
-bspline first, second, and third order derivatives
-band-limited-exponential(BLE) first and second order derivatives

These operators are the fruition of work from the MRA camp held at Stony Brook in 2016, and many of us are authors on the paper (submitted to Journal of Computational Physics) detailing their properties.

An example (derivative.cc) has been added to the examples directory showing how to use the new derivatives.

Keywords "deriv" and "deriv_dft" were added to CalculationParameters.h in the chem/ directory that allow for the new derivatives to be used inside moldft calculations.

Finally, a CMake variable (MAD_ROOT_DIR) that points to the MADNESS root directory was added to the top CMakeLists.txt file in order to allow for the use of these operators outside of the MRA directory.


fbischoff:
It's nice to have the new derivatives finally in the code. I've just played around a little and found the following:
 - derivative.cc includes "malloc.h", which breaks compilation, it can just be removed
 - SCF.cc:1521: in the parameter list for the XCOperator param.deriv is passed instead of param.dft_deriv.
 - CalculationParameters.h: There should be a check if the derivative method is actually valid

It would be great if we could have a derivative operator the that does not rely on vanishing or periodic functions at the boundaries, i.e. if at the boundary boxed not the central differences stencil would be used but a forward or backward differences stencil. Would that be complicated to implement?

PR #310, opened 2019-02-08
Mad nwinterface

bsundahl:
Adding into moldft the capability to restart from an NWChem output and movecs file. This will allow for moldft to calculate using any atom in the periodic table that NWChem has a full electron basis for. 

Added keyword "nwfile" to CalculationParameters.h, which facilitates the the interface.

This interface is largely derived from Slymer, an electron transport code being designed by mgr522 (currently still private).


robertjharrison (approved):
Nice job!   The code is better than when you found it.

PR #316, opened 2019-09-15
download from Gitlab because tddft.org is unreliable

jeffhammond:
Signed-off-by: Jeff Hammond <jeff.science@gmail.com>


evaleev:
'all' tests error due to excessive length, will create a separate issue

PR #319, opened 2019-10-18
PR: PNO-MP2

kottmanj:
Code for pno based MP2/CIS(D) calculations

The code has a lot of ways to construct guesses for PNOs left (like the pseudo LCAO construction) which are not described in the paper. I left them in, since they might come in handy at some point.
For that I needed to include boost into madness.
It's only used at one point and it seems like it works.
I integrated it similar to other external packages and it can be activated with cmake -D ENABLE_BOOST=ON (default is off).


kottmanj:
deleted lrccs in examples

PR #321, opened 2019-11-14
implementing replicated distributed containers

fbischoff:
implemented replicated functions for use in the construction of 6D functions from 3D functions.

The ProcessMap for each replicated function is generated uniquely in the replicate() method, please comment if that is consistent with the rest of world. A replicated function can be redistributed again, which will communicate all replicated boxes again (all-to-one). Another option would be to simply delete all boxes that are not local in the new map, but since I cannot say if a function is replicated that seems dangerous.


fbischoff:
Implemented a multi-world heterogeneous task-queue for large tasks that use subworlds, e.g. for computing parts of the Fock matrix. Relevant new files are parallel_dc_archive.h, macrotaskq.h, cloud.h, macrotaskpartitioner.h, test_vectormacrotask.cc

The implementation includes cloud.h for transferring data, including (pointers to) WorldObjects, from one world to another, the macrotaskq and corresponding base classes for the generation of the macrotasks itself.

Scaling with respect to the number of MPI processes is much improved, as tested for the exchange operator.

@Ed: please have a look at parallel_archive.h, line 400-440 where I've worked on your most recent commits

robertjharrison (approved):

PR #331, opened 2020-03-17
partial ASLR fix

evaleev:
along the lines of upc++ ... serialize fn ptrs relative to an origin ... must compile all user code into a single shared library to be able to avoid ASLR issues, so this is not a complete solution.


evaleev:
@robertjharrison this is good to go. Only tested on x86 (i.e. Itanium ABI), will need to test on ARM ... if you have a way to do that would be helpful.

PR #332, opened 2020-03-18
Transition to new parsec

evaleev:


evaleev:
@robertjharrison this is good to go

PR #339, opened 2020-04-29
Pr exe link change

robertjharrison:
1) arm port (minor changes except to mxm.h that was assuming MKL for fast blas)
2) use add_mad_executable for building all executables
3) fix protocol problem resulting from shift by Florian to CalculationParameters and then fix QC test programs to make them work and then shrink the tests since they were taking way too long

The add_mad_exe stuff was tested on ARM (using multiple MPI processes, release/debug builds, and shared libs on and off), and to a lesser extent on x86.


robertjharrison:
Forgot to include in the list of items the disabling of TBB by default.

PR #343, opened 2020-05-01
Selective build

keceli:
This PR might be very premature as I am still learning about CMake and MADNESS, but the goal is clear, to simplify the build process for users that are interested in only specific libraries of MADNESS. Added CMake options:
```
option(ENABLE_APPS "Enable building src/apps" ON)
option(ENABLE_EXAMPLES "Enable building src/examples" ON)
option(ENABLE_LAPACK "Enable using blas/lapack" ON)
option(ENABLE_WORLD_ONLY "Build only MADworld" OFF)
option(ENABLE_WORLD_CLAPACK_ONLY "Build MADworld and CLAPACK headers" OFF)
option(ENABLE_TENSOR_ONLY "Build all except mra" OFF)
```
With the default options , there is no change to the build system, but one can reduce the build time significantly using these options.


evaleev:
there already targets for building and installing individual libraries. Why are the `enable_x` cmake variables necessary?

keceli:
I couldn't find a simple way to build just a single library, so I thought having these options will help other users, too.

evaleev:
@keceli just use `build-x` and `install-x` to build and install library `x` and its dependents

evaleev:
@keceli OK, I'm slow but I think I understand the issue now. How about we only distinguish 2 cases: where only `MADWorld` (the runtime) is needed, and where `MADNESS` (`MADWorld` + other libs + potentially apps) is needed. So I suggest we:
- rename `ENABLE_WORLD_ONLY` -> `MADNESS_BUILD_MADWORLD_ONLY` (default is OFF)
- remove the rest of options
- do not create `madness` target when `MADNESS_BUILD_MADWORLD_ONLY=ON`

keceli:
@evaleev I'll change the option name as you suggested, but I'd like to find out how to proceed with other options.

We want the MADNESS built by SDE to be used for LibChemist or any other module that requires both SDE and TA. Since TA requires cblas/clapack header files in addition to MADworld, I added an option to install these header files, too. Otherwise, either LibChemist build its own MADNESS, or the build fails when it doesn't find the header files in SDE built MADNESS. Ideally, it should build only the missing files, but I couldn't find a way to do it. So, I added `ENABLE_WORLD_CLAPACK_ONLY` option, which builds MADworld and the required header files.

I also added `ENABLE_LAPACK` option to avoid searching for BLAS/LAPACK libraries, since for any build MADNESS searches for these libraries and fail if they are not found. However, these libraries are not required for SDE. If there is another way to avoid the search for BLAS/LAPACK, we can remove this option, too.

evaleev:
@keceli so SDE only wants to deal with MADNESS but wants for any TA-dependent code be able to use its copy of MADNESS? @ryanmrichard Would it be easier to just consume MADNESS via TA in SDE? This is going to be quite a bit of work ... much of the logic in TA's madness.cmake would have to be duplicated in SDE (and then maintained on the ongoing basis). I agree that the right solution is to actually modularize MADNESS' CMake harness properly but even then the logic of configuring MADNESS correctly remains.

ryanmrichard:
@evaleev that's correct, the idea is to decouple the runtime, which is all SDE needs. My guess is that TA (and anything else that uses MADNESS) will need to be able to detect individual components so that we don't end up linking in two different MADNESS runtimes. 

It would be easier for SDE to consume TA, but design wise we don't want that. SDE is supposed to be tensor/math agnostic, which should aid in interoperability with other codes. It also facilitates switching between TA and TAMM (in practice much of the software stack built on SDE is going to need to change for this to occur, but SDE won't need to). 

Admittedly the complexity is part of the reason why I have been putting off adding MADNESS to SDE. At the moment almost everyone is developing in LibChemist (where TA enters the stack) or downstream of it, so they have access to MADNESS through it. Admittedly we're getting to the point where that can't be ignored too much longer as the Cache is going to need to be aware of the parallel and resources available.

keceli:
@evaleev I did the changes you requested, but still kept `MADNESS_BUILD_MADWORLD_CLAPACK_ONLY` and `MADNESS_SEARCH_LAPACK` options. Former is required to add CLAPACK headers for TA compatibility and the latter is required to turn off the search for BLAS/LAPACK.  
- rename `ENABLE_WORLD_ONLY` -> `MADNESS_BUILD_MADWORLD_ONLY` (default is OFF)
- rename `ENABLE_WORLD_CLAPACK_ONLY` -> `MADNESS_BUILD_MADWORLD_CLAPACK_ONLY` (default is OFF)
- rename `ENABLE_LAPACK` -> `MADNESS_SEARCH_LAPACK` (default is ON)
- remove the rest of options
- do not create madness target when `MADNESS_BUILD_MADWORLD_ONLY=ON`

evaleev:
@keceli please review, test with SDE, and address any travis issues (if any)

PR #345, opened 2020-05-07
Pr mad dq prebuf

robertjharrison:
To nearly eliminate the performance gap between the threadpool and TBB I added a thread-local pre-buffer to aggregate tasks to reduce lock contention.  Tuning done on quadsocket intel box for (h2o)5.  In a separate commit also introduced slight symmetry breaking to make localized orbitals better defined for molecules with C3 or higher.

(H2O)5 LDA timings
* run on sn-mem (72 core, quad socket, Skylake)
* all with tbbmalloc preloaded
* threads were bound to a socket unless noted otherwise

Option          P    T   time(s)                 comment
MAD+PRE    4   16   36.5,37.6,41.8,47.5
MAD+PRE    1   16   114.9,116.1,119.4
MAD+PRE    1   16   120.0,119.7            ninprebuf=10 (default is 20)
MAD+PRE    1   16   117.8                  ninprebuf=40 (default is 20)
MAD+PRE    1    8   148.1,161
MAD+PRE    1    4   265.2
MAD+PRE    1    2   420.7,424.1

MAD+PRE    1    2   433                    nobind
MAD+PRE    1   16   119.4                  nobind
MAD+PRE    1   16   112.8,123.1            nobind ninprebuf=10 (default is 20)
MAD+PRE    1   16   109.4                  nobind ninprebuf=40 (default is 20)
MAD+PRE    1   32   126.0                  nobind ninprebuf=10 (default is 20)
MAD+PRE    1   32   119.0                  nobind
MAD+PRE    1   32   108.0,110.0            nobind ninprebuf=40 (default is 20)
MAD+PRE    1   64   157.2                  nobind
MAD+PRE    1   64   131.5                  nobind ninprebuf=40 (default is 20)

MAD+PRE    2    1   334.6

MAD+PRE    1    1   478

MAD        4   16   48.6,50.9,54.8
MAD        1   16   148.9,149.4
MAD        1    8   164.5

TBB        4   16   37.0,38.2
TBB        4    8   55.4
TBB        1   16   99.0,99.4
TBB        1    8   140.9
TBB        1    2   414.8

TBB        1    2   401.3                     nobind
TBB        1   16   95.7                      nobind
TBB        1   32   77.1                      nobind
TBB        1   64   87.8                      nobind


CMakeLists.txt
```diff
@@ -552,7 +559,7 @@ endif(CCACHE)
 
 # Check for external dependencies ==============================================
 include(external/pthread.cmake)
-include(external/mpi.cmake)
+#include(external/mpi.cmake)
```
evaleev, line 562: was this meant to be commented out?

src/madness/world/dqueue.h
```diff
@@ -78,9 +85,18 @@ namespace madness {
         volatile int _back;    ///< Index of element at back of buffer
         DQStats stats;
 
+#ifdef MADNESS_DQ_USE_PREBUF
+	static const size_t NPREBUF=20;
+	static __thread T prebuf[NPREBUF]; // relies on this being a singleton class!!!!!!!!!!!!!!!!!!
```
evaleev, line 90: to be portable need to use `thread_local` ... https://en.cppreference.com/w/cpp/keyword/thread_local

CMakeLists.txt
robertjharrison: Good catch ... on my machine mpi.cmake just gets stuck in an infinite (or at least 5min+) long loop so I just commented it out --- everything seems to work fine without it.

I'll repush with it uncommented in case it is needed on other platforms.

src/madness/world/dqueue.h
robertjharrison: Okay.

src/madness/world/dqueue.h
```diff
@@ -78,9 +85,18 @@ namespace madness {
         volatile int _back;    ///< Index of element at back of buffer
         DQStats stats;
 
+#ifdef MADNESS_DQ_USE_PREBUF
+	static const size_t NPREBUF=20;
```
evaleev, line 89: presumably the optimal is application (or better yet op-dependent) so do we want runtime control of the granularity? Our tasks are fairly fat so I'd imagine we don't want this param as high ... at least make it a CMake-configurable value

src/madness/world/worldrmi.cc
```diff
@@ -165,6 +165,12 @@ namespace madness {
 
             post_pending_huge_msg();
 
+#if !(defined(HAVE_INTEL_TBB) || defined(HAVE_PARSEC))
+	    // Since this thread never waits or tries to run a task we
+	    // must manually flush the thread-local prebuffer that
+	    // aggregates task submission.
+	    ThreadPool::instance()->flush_prebuf();
```
evaleev, line 172: tabs-vs-spaces debates aside (https://thenewstack.io/spaces-vs-tabs-a-20-year-debate-and-now-this-what-the-hell-is-wrong-with-go/) I think most code in MADNESS does not use tabs ... looks like a tab here and elsewhere?

src/madness/world/dqueue.h
robertjharrison: Making this runtime configurable will slow down the code so I left it is compile time configurable.

src/madness/world/worldrmi.cc
robertjharrison: Dunno where the tabs came from ... just untabified the whole file.

src/madness/world/dqueue.h
evaleev: Due to the need to fetch a line to compare ninprebuf{,hi} to this?

OK, can be addressed later.

PR #349, opened 2020-05-26
Remove multilib support

evaleev:


evaleev:
@robertjharrison switching back to single-lib build ... simpler and there is no consensus on what needs to be best done to support multilib (shared+static) for packageability etc. See INSTALL for the discussion of how to deal with ASLR.

PR #353, opened 2020-07-21
Templated wrapper interface for select LAPACK routines

wavefunction91:
Currently, there is only an "internal use" implementation of template wrappers for LAPACK routines interfaced to raw pointers with the correct LAPACK/LAPACKE function wrappers. This PR exports a templated API which delegates to the proper LAPACK/LAPACK wrappers and can be used downstream.

This PR adds template wrappers for the following LAPACK routines:
- XGESVD
- X{SY,HE}EV
- X{SY,HE}GV
- XPOTRF

### DIFF
--- a/src/madness/tensor/CMakeLists.txt
+++ b/src/madness/tensor/CMakeLists.txt
@@ -30,7 +30,7 @@ set(MADTENSOR_SOURCES tensor.cc tensoriter.cc basetensor.cc vmath.cc)
 # Source lists for MADclapack
 set(MADCLAPACK_HEADERS cblas.h cblas_types.h
         clapack.h clapack_fortran.h
-        lapacke_types.h) # this part of MADlinalg is purely independent of MADtensor
+        lapacke_types.h linalg_wrappers.h ) # this part of MADlinalg is purely independent of MADtensor
 
 # Source lists for MADlinalg
 set(MADLINALG_HEADERS ${MADCLAPACK_HEADERS} tensor_lapack.h solvers.h elem.h)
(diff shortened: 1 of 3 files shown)


wavefunction91:
Ping @evaleev

robertjharrison:
I thought nwchemex was only using the world APIs from madness and so was never building anything else.  

I am not in any way opposed to other APIs being used, but it is a slippery slope.  Perhaps the right thing to do is to factor what you want out into a separate directory in madness instead having to build the tensor directory.   As previously conceived all of the madness linear algebra was exposed only thru our tensor class which you probably don't want people using.

evaleev:
@robertjharrison this is to leverage the BLAS/LAPACK/CBLAS/LAPACKe interface only, not libMADtensor. Since BLAS/LAPACK API lives in global namespace one has to (rather: should) use the same interface across the entire app. TA already uses this, so this issue infects the entire NWX stack.

Yes, factoring out the whole shared-memory linear algebra into its own package makes sense, but then it probably also makes sense to provide proper cmake harness for it (exported targets, etc.). We keep hacking because doing the right thing is too much work and ICL is not too interested.

wavefunction91:
I agree, we should probably factor this out into a proper `MADlinalg` (or something). The reason this needs to get propagated downstream simultaneously to World is that TA delegates BLAS/LAPACK linkage to MAD.

The correct thing to do would be to include the `blaspp`/`lapackpp` wrappers from the `SLATE` project (as @evaleev mentioned), but their build system leaves much to be desired (building them as subprojects is nontrivial). It might be worth forking and fixing them, but that would be a lot of work and there's no gurentee that they would accept changes upstream.

I'm more than happy to factor these into a separate directory, but that would also require factoring out the LAPACK/LAPACKE bindings as well. You can see how this might spiral pretty quickly, especially if we include the BLAS/CBLAS bindings as well.

PR #355, opened 2020-08-03
fence takes callback to execute prior to release

evaleev:
currently fence does not suspend the pool, so this does not do anything useful, this is only to start a discussion


src/madness/world/worldgop.cc
```diff
@@ -142,6 +142,8 @@ namespace madness {
             nrecv_prev = sum[1];
 
         };
+        // execute post-fence actions ... do we want to execute these with pool paused?
+        pre();
```
evaleev, line 146: @robertjharrison to execute user-provided and implicit (cleanup) actions safely we should pause the thread pool. Do you agree?

PR #356, opened 2020-08-04
fix futures

robertjharrison:
Not ready yet --- in addition to the reference problem there seems to be an issue associated with using futures for the results of remote tasks (but not remote active messages?????).

### DIFF
--- a/src/madness/world/future.h
+++ b/src/madness/world/future.h
@@ -38,6 +38,12 @@
 #ifndef MADNESS_WORLD_FUTURE_H__INCLUDED
 #define MADNESS_WORLD_FUTURE_H__INCLUDED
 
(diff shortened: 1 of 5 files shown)


robertjharrison:
Time for you to take a look

src/madness/world/future.h
```diff
@@ -275,7 +287,7 @@ namespace madness {
             //
             // Also assume that the caller either has the lock
             // or is sure that we are single threaded.
-            MADNESS_ASSERT(!assigned);
+            MADNESS_CHECK(!assigned);
```
evaleev, line 290: what's the rationale for `CHECK` rather than `ASSERT`, here and elsewhere?

PR #358, opened 2020-08-18
amends #356 to prevent unsafe move-out

evaleev:


evaleev:
@robertjharrison the rvalue-ref Future::get in #356 was not safe since the fact that sole ownership is what matters here ... unfortunately requires `std::atomic_store<std::shared_ptr>` whose efficiency is unknown, it's also deprecated in C++20 (`Future::f` would need to be converted to `std::atomic<std::shared_ptr<FutureImpl>>` to support atomic ops in C++20+)

evaleev:
bump @robertjharrison

robertjharrison (approved):
Sorry ... thought this was already merged.

After some initial reservations I agree with the logic for moving.  We might eventually want to separate the two uses of futures --- a final variable that only ever has one reader from a holder of data that might have multiple readers. 

Once the tests are run I will merge.

evaleev:
> 
> After some initial reservations I agree with the logic for moving. We might eventually want to separate the two uses of futures --- a final variable that only ever has one reader from a holder of data that might have multiple readers.

That's basically why there is `std::future` and `std::shared_future`. `madness::Future` is a composable `std::shared_future`

PR #359, opened 2020-09-16
Pr oep

fbischoff:
finalized code for the computation of optimized effective potentials -- should be mostly orthogonal to the rest of the code


robertjharrison (approved):
Good to go!

PR #361, opened 2020-10-07
uniqueidT is only explicitly convertible to bool (implicit conversion…

evaleev:
… allowed all kinds of nonsense to compile, e.g. uniqueidT < uniqueidT) + added inequality and ordering operators


evaleev:
nudge @robertjharrison

PR #362, opened 2020-10-19
constrain broadcast_serializable to serializable types +

evaleev:
+ make archive_array<T> serializable if T is


evaleev:
@robertjharrison looks like some types that should not be serializable are broadcast as such: https://travis-ci.com/github/m-a-d-n-e-s-s/madness/jobs/402215842#L6211 ... probably they are PODs that are not trivially copyable ... but this is dangerous enough that I'd say disabling for nonserializable types is a good idea.

PR #367, opened 2021-02-03
Removing of deprecated tbb::task API

anton-potapov:
This patch 
 - replaces redundant overrides of `tbb::task::execute` method with overrides of `TaskInterface::run(World&, const TaskThreadEnv&)`
 - added explicit `tbb::task_arena` to be used instead of implicit one
 - `PoolTaskInterface` migrated from `tbb::task` usage 
 - `MADNESS_CAN_USE_TBB_PRIORITY` is disabled at the moment


src/madness/world/thread.h
```diff
@@ -1328,12 +1302,15 @@ namespace madness {
...
-#endif  // MADNESS_CAN_USE_TBB_PRIORITY
-                tbb::task::enqueue(*task);
+//#ifdef MADNESS_CAN_USE_TBB_PRIORITY
+//            if(task->is_high_priority())
+//                tbb::task::enqueue(*task, tbb::priority_high);
+//            else
+//#endif  // MADNESS_CAN_USE_TBB_PRIORITY
+            tbb_arena->enqueue(
+                    [task_p = std::unique_ptr<PoolTaskInterface>(task)](){
+                        task_p->execute();
```
alexey-katranov, line 1312: It seems you copied a pointer of `PoolTaskInterface` from `task` to `task_p` via `unique_ptr`. Why not just use `task` directly, e.g.
```suggestion
                    [task](){
                        task->execute();
```

anton-potapov: it has to be deleted as well, which is for `unique_ptr` is used

alexey-katranov: So, you are using ‘unique_ptr’ as an raii guard while an exception is not expected (it will lead to UB if thrown from enqueue task). Does it make sense to specify ‘no except’ for lambda to cause termination in case of exception? I’m addition, add a comment that ‘unique_ptr’ is used to deallocate ‘task’ after execution.

anton-potapov:
@evaleev ,Could you please tell me -  should I keep those commits you pushed in , or they can be removed ?

evaleev:
@anton-potapov yes, please keep them in. I'm still doing more tests of your changes.

src/madness/world/thread.h
```diff
@@ -1377,20 +1356,6 @@ namespace madness {
...
-
-            // Create a dummy task that we send throught the queue.
-            tbb::task& dummy = *new( waiter.allocate_child() ) tbb::empty_task;
-            tbb::task::enqueue(dummy);
-
-            // Run tasks while we wait for the dummy task.
-            waiter.wait_for_all();
-
-            // destroy the waiter
-            tbb::task::destroy(waiter);
```
alexey-katranov, line 1393: Is it some sort of moonlighting? We can try to implement similar approach with `task_hadnle`; however, the overall idea seems fragile.

src/madness/world/worldrmi.cc
```diff
@@ -401,56 +399,20 @@ namespace madness {
...
-#else
-                tbb::task::enqueue(*empty);
-#endif  // MADNESS_CAN_USE_TBB_PRIORITY
-                ++binge_counter;
-              }
               myusleep(100000);
             }
-            empty_root->wait_for_all();
-            tbb::task::destroy(*empty_root);
-            task_ptr->comm.Barrier();
```
alexey-katranov, line 451: What about if worker threads are not available? Previously, `RmiTask` can be picked up with the current thread.

robertjharrison:
Early TBB design allowed main thread to execute tasks but that was removed.
So there will now always be a main thread and a worker thread.

Then it seemed that TBB would not permit other threads to be created
(!why?) so we had to force the RMI thread into a task but TBB was lazy
starting threads hence the stupid pushing of empty tasks and probing to
ensure the RMI task is running.  Same thing here ... How to make sure the
queue is really empty. No way to just ask.

I don't know the term ”moonlighting" except in the sense of dual purpose.

src/madness/world/thread.h
anton-potapov: no

alexey-katranov:
@robertjharrison 

> Early TBB design allowed main thread to execute tasks but that was removed.

The main thread is allowed to join parallelism when waiting for work completion. Do you mean that the main thread does execute the tasks when finishes?

> Then it seemed that TBB would not permit other threads to be created

It seems really strange, TBB tries to utilize all the resources allowed by the process affinity mask. So, TBB will not create threads if there is only one bit in process affinity mask.

> ”moonlighting"

We usually use this term to describe the behavior when thread executes something that does not related the waiting condition. In other words it is doing some additional work that is not expected by its wait condition.

anton-potapov:
@robertjharrison  @evaleev can we move this forward ?

PR #368, opened 2021-02-05
Fix madness::PNO::get_average_rank() on 32bit

AdrianBunk:
On 32bit size_t is not the same as unsigned long int, causing:

https://buildd.debian.org/status/fetch.php?pkg=madness&arch=i386&ver=0.10.1%2Bgit20200818.eee5fd9f-1&stamp=1609527654&raw=0

...
/<<PKGBUILDDIR>>/src/apps/chem/PNO.cpp:632:49: error: no declaration matches ‘std::pair<long unsigned int, long unsigned int> madness::PNO::get_average_rank(const std::valarray<std::vector<madness::Function<double, 3> > >&) const’
  632 | std::pair<unsigned long int, unsigned long int> PNO::get_average_rank(const std::valarray<vector_real_function_3d>& va) const {
      |                                                 ^~~
In file included from /<<PKGBUILDDIR>>/src/apps/chem/PNO.cpp:8:
/<<PKGBUILDDIR>>/src/apps/chem/PNO.h:96:28: note: candidate is: ‘std::pair<unsigned int, unsigned int> madness::PNO::get_average_rank(const std::valarray<std::vector<madness::Function<double, 3> > >&) const’
   96 |  std::pair<size_t, size_t> get_average_rank(const std::valarray<vector_real_function_3d>& va) const;
      |                            ^~~~~~~~~~~~~~~~


kottmanj (approved):
Looks good. Thanks!

PR #373, opened 2021-05-08
serialization sfinae

evaleev:
- instantiations are protected by SFINAE to be able to detect serializability of a type correctly.
- user-provided serialization methods override default serialization for trivially-serializable methods

top level operators (`&`, `<<`, `>>`) are unprotected but static_assert serializability inside to produce more reasonable error messages.


robertjharrison:
Nice!  Will merge once CI completes.  Won't be able to play with it myself until late tonight.

PR #374, opened 2021-05-11
madness serialization cleanup

evaleev:
this reverts the support for versioned variants of serialize introduced in #373 ... it's safer not to try to mimic interoperability with boost.

many more sfinae fortifications + removed some ambiguous overloads (they became  ambiguous *enough* after sfinae for the compiler to complain)


evaleev:
@robertjharrison this is good to review/merge

PR #375, opened 2021-05-27
Added ENABLE_CEREAL option with default OFF

keceli:
When this option is on, cereal is downloaded and installed.


CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ set(MADNESS_INSTALL_CMAKEDIR "lib/cmake/madness"
 # Build options  ==============================================================
 
 redefaultable_option(MADNESS_BUILD_MADWORLD_ONLY "Build only MADWorld runtime (excluding BLAS/LAPACK interfaces)" OFF)
+redefaultable_option(ENABLE_CEREAL "Use Cereal library along with MADNESS archive for serialization" OFF)
```
evaleev, line 69: should rename to make it MADNESS-specific (`MADNESS_ENABLE_CEREAL`)

cmake/modules/FindCereal.cmake
```diff
@@ -0,0 +1,19 @@
+find_package( cereal QUIET )
+if( cereal_FOUND )
+    target_compile_definitions(cereal INTERFACE
+      "CEREAL_THREAD_SAFE=1;GAUXC_ENABLE_CEREAL=1;JUST_INSTALL_CEREAL=ON;SKIP_PORTABILITY_TEST=ON")
+else()
+    message( STATUS "Could not find Cereal. Downloading..." )
+    include(FetchContent)
+    FetchContent_Declare(
+      cereal
+      GIT_REPOSITORY git@github.com:USCiLab/cereal.git)
```
evaleev, line 10: need to pin to specific commit

evaleev (changes requested):
since now MADNESS tests for presence of cereal should define macro and use that to enable cereal-based archives rather than use `__has_include` (the latter can trigger even if `MADNESS_ENABLE_CEREAL=OFF`)

cmake/modules/FindCereal.cmake
keceli: Done with 7412c09.

CMakeLists.txt
keceli: Done with 7412c09.

keceli:
> since now MADNESS tests for presence of cereal should define macro and use that to enable cereal-based archives rather than use `__has_include` (the latter can trigger even if `MADNESS_ENABLE_CEREAL=OFF`)

Thank you @evaleev for reviewing. I added `add_compile_definitions(MADNESS_HAS_CEREAL)` to `FindCereal.cmake` and set an error if `MADNESS_ENABLE_CEREAL=ON` but cereal header is not found.

cmake/modules/FindCereal.cmake
```diff
@@ -0,0 +1,22 @@
...
+    FetchContent_GetProperties(cereal)
+    if(NOT cereal_POPULATED)
+        FetchContent_Populate( cereal )
+        add_library( cereal INTERFACE IMPORTED )
+        set_target_properties( cereal PROPERTIES
+          INTERFACE_INCLUDE_DIRECTORIES "${cereal_SOURCE_DIR}/include"
+          INTERFACE_COMPILE_DEFINITIONS "CEREAL_THREAD_SAFE=1;GAUXC_ENABLE_CEREAL=1;JUST_INSTALL_CEREAL=ON;SKIP_PORTABILITY_TEST=ON")
+  endif()
+endif()
+add_compile_definitions(MADNESS_HAS_CEREAL)
```
evaleev, line 21: old-school global definitions like these should be avoided in modern CMake. Either needs to go to madness_config.h or better defined for MADworld and madness targets

evaleev (changes requested):

cmake/modules/FindCereal.cmake
keceli: Would this abb7c1c2 be fine?

cmake/modules/FindCereal.cmake
```diff
@@ -1,20 +0,0 @@
...
-      cereal
-      GIT_REPOSITORY git@github.com:USCiLab/cereal.git
-      GIT_TAG v1.3.0)
-    FetchContent_GetProperties(cereal)
-    if(NOT cereal_POPULATED)
-        FetchContent_Populate( cereal )
-        add_library( cereal INTERFACE IMPORTED )
-        set_target_properties( cereal PROPERTIES
-          INTERFACE_INCLUDE_DIRECTORIES "${cereal_SOURCE_DIR}/include"
-          INTERFACE_COMPILE_DEFINITIONS "CEREAL_THREAD_SAFE=1;GAUXC_ENABLE_CEREAL=1;JUST_INSTALL_CEREAL=ON;SKIP_PORTABILITY_TEST=ON")
```
evaleev: @keceli it's not Cereal's find module to be aware of other modules. respect encapsulation. Gauxc needs to check for presence of cereal target and configure itself appropriately.

evaleev (approved):

evaleev:
@keceli after your changes CI never actually used cereal ... please fix, still not working for some reason

keceli:
> @keceli after your changes CI never actually used cereal ... please fix, still not working for some reason

7de6f6e should fix the problem, but reverts some of the changes you made in dbeaacb.

PR #401, opened 2021-10-30
Add serialization for std::set

keceli:


evaleev (approved):

keceli:
I have two questions. 
1. How do we enable CI for this PR? I thought it was automatic, but apparently not. 
2. There are many other `std` types that are not supported by MADNESS archive, but supported by Cereal. Is there a way to automatically support all the types Cereal supports for MADNESS archive, at least for non-parallel archives?

evaleev:
> 1. How do we enable CI for this PR? I thought it was automatic, but apparently not.

CI ran: https://github.com/m-a-d-n-e-s-s/madness/actions/runs/1546517196

> 2. There are many other `std` types that are not supported by MADNESS archive, but supported by Cereal. Is there a way to automatically support all the types Cereal supports for MADNESS archive, at least for non-parallel archives?

No. Just like Cereal, must specify how to serialize each `std` type manually.

evaleev:
> 2. There are many other `std` types that are not supported by MADNESS archive, but supported by Cereal. Is there a way to automatically support all the types Cereal supports for MADNESS archive, at least for non-parallel archives?

Which other types do you need?

keceli:
> Which other types do you need?

`std::shared_ptr`, `std::variant`, and probably for `std::unique_ptr`.

keceli:
> > 1. How do we enable CI for this PR? I thought it was automatic, but apparently not.
> 
> CI ran: https://github.com/m-a-d-n-e-s-s/madness/actions/runs/1546517196

There it says 12 jobs completed, but I still see continuous-integration/travis-ci "Expected — Waiting for status to be reported". How to start that one?

evaleev:
> There it says 12 jobs completed, but I still see continuous-integration/travis-ci "Expected — Waiting for status to be reported". How to start that one?

Travis CI is no longer used

evaleev:
> `std::shared_ptr`, `std::variant`, and probably for `std::unique_ptr`.

the latter 2 are easy. What do you actually want serialization of `std::shared_ptr` to do?

robertjharrison:
Need these to also make sense in distributed memory setting and to be
consistent with remote reference usage.

keceli:
> the latter 2 are easy. What do you actually want serialization of `std::shared_ptr` to do?

I need it to serialize/deserialize the [`CanonicalSpace`](https://github.com/NWChemEx-Project/Chemist/blob/master/include/chemist/orbital_space/canonical_space.hpp) instances within NWChemEx/Chemist.

evaleev:
> > the latter 2 are easy. What do you actually want serialization of `std::shared_ptr` to do?
> 
> I need it to serialize/deserialize the [`CanonicalSpace`](https://github.com/NWChemEx-Project/Chemist/blob/master/include/chemist/orbital_space/canonical_space.hpp) instances within NWChemEx/Chemist.

Do you need to be able to serialize/deserialize graphs of objects? I.e. if 2 objects hold `shared_ptr` to the same `CanonicalSpace` object do you expect these 2 objects, when deserialized, to refer to the same `CanonicalSpace` object still? To implement that you need object/pointer tracking. MADNESS serialization does not currently consider such scenario as a use case since the primary goal is to serialize simple data structures across memory spaces, not for persistence. It should be possible to provide archives intended for serializing object graphs, it might be easier though to do that by offloading that work to cereal/boost.

ryanmrichard:
@evaleev I would expect `std::shared_ptr` serialization to work the way you describe (if upon serialization pointers `a`, `b`, and `c` all point to the same object, then upon deserialization `a`, `b`, and `c` should all point to the same object). Unless you're using `std::shared_ptr` as a behind the scenes shallow copy optimization, I think most use cases rely on such behavior. FWIW, I'm pretty sure that's how the Boost and Cereal implementations already work, although it's been a long time since I've looked at them.

evaleev (approved):

PR #406, opened 2021-11-30
PaRSEC integration update

therault:
- cleanup integration: split parsec-related operations in a parsec class and corresponding files
- bump up the parsec version used: 
   - use parsec taskpool polling to allow mobilizing the main thread into the computation
   - use a version of parsec that does not interact with libc rand()
   - use a version of parsec that does not compile and run parsec tests related to PTG or DTD
   - remove warnings at compile time


src/madness/world/thread.h
```diff
@@ -1291,9 +1285,9 @@ namespace madness {
 	}
 
 #if HAVE_PARSEC
-	////////////////// Parsec Related Begin //////////////////
-        static parsec_context_t *parsec;
-        ///////////////// Parsec Related End ////////////////////
+	    ////////////////// Parsec Related Begin //////////////////
+	    static ParsecRuntime *parsec_runtime;
```
devreal, line 1289: Why is parsec_runtime static? It's deleted in the dtor of Taskpool (and not set to `null`) so it should be a member of the taskpool, no?

src/madness/world/parsec.h
```diff
@@ -15,12 +15,21 @@
...
-
-    extern const parsec_task_class_t madness_parsec_tc;
-    extern parsec_taskpool_t madness_parsec_tp;
-  }
-
+    class PoolTaskInterface;
+
+    class ParsecRuntime {
+    private:
+        static parsec_context_t *context;
```
devreal, line 22: Why are they all static? The parsec_runtime object is already static and even if it wasn't, I don't see how there can a constructor that intiializes static members, esp. since that constructor asserts if context is already set.

devreal (changes requested):
I'm confused about the static members in a few places. Do they really have to be static?

src/madness/world/parsec.cc
```diff
@@ -126,11 +112,76 @@ namespace madness {
...
             .repo_array = NULL
-        };
+    };
+    parsec_context_t *ParsecRuntime::context = nullptr;
+
+    ParsecRuntime::ParsecRuntime(int nb_threads) {
+        assert(context == nullptr);
+        /* Scheduler init*/
+        int argc = 1;
+        char ** argv = (char**)malloc(2*sizeof(char*));
```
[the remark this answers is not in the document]
therault: Will fix

src/madness/world/thread.h
therault: begin(), end(), and add(), that all call methods of this object, are all defined static, so this pulled all those to become static too.

The ParsecRuntime class holds a parsec_context_t and a parsec_taskpool_t, so I'm not sure I understand your question about the ParsecRuntime being a member of the taskpool.

src/madness/world/parsec.h
therault: schedule(), test() and task() are called from  static methods in ThreadPool, that makes them become static, and then they refer the taskpool, that needed to become static too.

devreal: Sorry, let me clarify: The ThreadPool has a static member parsec_runtime. There is no need for the member variables and functions of the ParsecRuntime to be static, too. They can be called like any member function on the static parsec_runtime member of ThreadPool. 

The other question would be: can there be more than one ThreadPool? And if so, who has ownership of the parsec context? (i.e., who is responsible for its destruction?) If there can be more than one over the lifetime of an application, the ctor ParsecRuntime would assert on the second creation of the ThreadPool...

evaleev: @devreal `ThreadPool` is a singleton. So its methods (in this particular design) became static functions

devreal (approved):

PR #460, opened 2023-02-06
check if MADCHEM has file system in molresponse

ahurta92:


src/apps/molresponse/ResponseBase.cpp
```diff
@@ -1265,32 +1294,35 @@ void ResponseBase::plotResponseOrbitals(World &world, size_t iteration,
         plotCoords plt(d, Lp);
         // plot ground density
         if (iteration == 1) {
-            snprintf(plot_name, buffSize, "plots/densities/rho0_%c_0.plt", dir[d]);
+            auto d_i_path = density_dir + "/rho0_%c_0.plot";
```
fbischoff, line 1297: if you don't have std::filesystem, will this line create the file in the root directory?

ahurta92: @fbischoff  Good point.  I wasn't sure, so I did some testing.   I've made some changes, and everything works fine.  The code writes to the working directory if  `std::filesystem` does not exist. Feel free to let me know if there is anything else I should change.  Thanks

fbischoff (approved):

PR #477, opened 2023-05-31
Molresponse development

ahurta92:


src/madness/chem/SCF.cc
ahurta92, line 1: Contains some edits to `output_scf_info_schema` and `output_calc_info_schema` also writes dipole moment to `output_scf_info`

src/madness/chem/molecule.h
ahurta92, line 1: Adds the Molecule::to_json declaration

src/apps/moldft/moldft.cc
ahurta92, line 1: Only world 0 writes to the json file

src/apps/molresponse/CMakeLists.txt
ahurta92, line 1: molresponse has been restructured significantly.  I've removed several files that are no longer used

PR #484, opened 2023-06-12
adding plot apps: plot2plane and plot2cube

kottmanj:
more convenient plotting of standard functions produced by nemo, pno, moldft
easier to find than examples/tiny.cpp

### DIFF
--- a/src/apps/CMakeLists.txt
+++ b/src/apps/CMakeLists.txt
@@ -1,7 +1,8 @@
 # src/apps
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
-  
+
+add_subdirectory(plot)
 add_subdirectory(tdse)
 add_subdirectory(moldft)
 add_subdirectory(molresponse)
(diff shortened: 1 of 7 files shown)


fbischoff:
plot2plane doesn't seem to work well, esp. the determination of the box, where in CalculationParameters.h 50 bohr are added, but not in plot2plane or also in plot2cube. Maybe it would be worth an idea to add the box size into the serialization of a Function?

Generally I would use CalculationParameters, to unify and simplify things. That would give the highest probability of having an input consistent with moldft.

Suggestion: have multiple files plotted, that makes it easier in gnuplot to compare these functions directly.

Is there a reason that in CMakeLists.txt only plot2cube will be installed, but not plot2plane?

evaleev:
@fbischoff is this ready to go? lg2m

fbischoff:
Looks ok to me. A nice feature would be to do the same thing without the input file, but that will require the box size in the Function, so let's leave this for later

fbischoff:
Looks ok to me. A nice feature would be to do the same thing without the input file, but that will require the box size in the Function, so let's leave this for later

fbischoff (approved):
good to go

PR #490, opened 2023-07-18
Pr cis restart

fbischoff:


fbischoff:
This is pr is mostly concerned with restart options, output formatting and making usage more consistent, e.g. the option "excitations" was used differently in "response" and in "cc2". More changes will come, but this seems a good checkpoint

PR #512, opened 2023-12-28
harden World unique object handling

evaleev:
`World::id_to_ptr()` return an optional; if `NDEBUG` is not defined then its value for deregistered IDs is nullptr, this can be used to detect use of unregistered/deleted objects.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -33,6 +33,7 @@ jobs:
       CTEST_OUTPUT_ON_FAILURE : 1
       BUILD_CONFIG : >
         -G Ninja
+        -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
         -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/install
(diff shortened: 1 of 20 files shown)


robertjharrison:
Interesting ... what's driving this?

evaleev:
> Interesting ... what's driving this?

larger jobs (with hundreds of thousands of expression elements, involving thousands of tiles each) are getting stuck stochastically (but with near 100% probability) due to lack of progress. I thought maybe a lifetime issue (pulling from temporary objects, e.g. subexpressions), but does not seem to be that. Expression evaluation actually pushes data up, which makes lifetime management easier. TA expression cleanup does not use global fencing, only local info, and push makes it possible.

evaleev:
@robertjharrison I don't think I made any perf-related changes here, but if you see something amiss let me know.

robertjharrison:
Is there an issue with ordering or tasks internally forcing futures?  Or
use of send instead of task?

Grasping at straws in the absence of info.

evaleev (commented):
approved

PR #520, opened 2024-01-23
introduced `WorldGopInterface::{set_,}max_reducebcast_msg_size()`

evaleev:
They control the max size of buffers used by reduce/broadcast ... this may help the issues with MPI errors seen on Perlmutter when sum-reducing large arrays.

Control programmatically (`world.gop.set_max_reducebcast_msg_size()`) as needed or once at startup by setting environment variable `MAD_MAX_REDUCEBCAST_MSG_SIZE`

### DIFF
--- a/src/madness/world/safempi.h
+++ b/src/madness/world/safempi.h
@@ -533,6 +533,13 @@ namespace SafeMPI {
                 return result;
             }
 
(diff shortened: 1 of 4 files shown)


evaleev:
might be related to the issue this PR tries to work around: https://github.com/open-mpi/ompi/issues/6091

PR #523, opened 2024-02-06
memory size units

evaleev:
add literals for memory size (e.g. `_MiB` and `_MB`) and unify conversion of string values of environment variables `MAD_{BUFFER,MAX_REDUCEBCAST_MSG}_SIZE` to # of bytes


evaleev:
@robertjharrison this is ready to go ... since master is broken merging with 🤞

PR #530, opened 2024-03-06
Fixes problem with setting units to au or Bohr

ahurta92:
If using au or bohr the set_derived_value would fail when type checking bare "atomic" in function call.  std::string("atomic") ensures type checking will not fail.


JonathonMisiewicz:
@ahurta92 Rebase onto master so tests pass.

devreal:
@ahurta92 Is this still relevant or can we close the PR?

ahurta92:
Pretty sure the change got included in another PR

PR #531, opened 2024-03-13
Misc. cleanup

JonathonMisiewicz:
Lots of misc. code modernization. Range-based for loops, `make_shared` instead of `new`, `auto`, and structured bindings.


JonathonMisiewicz:
This PR and #530 both show the same two tests stuck in the `Install` phase, so I can say with confidence that something isn't correct in the test setup. Is this a known problem?

evaleev:
@JonathonMisiewicz please rebase on master

evaleev (approved):

PR #545, opened 2024-08-19
Pr fix cc2

fbischoff:
merging cc2 branch back into master -- all doubles equations are now running in macrotasks, local orbitals work


fbischoff:
only the gnuplot test fails, so I proceed with merging

PR #547, opened 2024-09-12
Decrease time for world.gop.reduce

EricaCMitchell:
Decreases time for world.gop.reduce by shortening the buffer length to the minimum number of elements and only forming the buffers if the child process exists. Significantly, speeds up sum reduction for `std::complex` type.
@evaleev

### DIFF
--- a/src/madness/world/worldgop.h
+++ b/src/madness/world/worldgop.h
@@ -785,28 +785,28 @@ namespace madness {
           world_.mpi.binary_tree_info(0, parent, child0, child1);
           const std::size_t nelem_per_maxmsg = max_reducebcast_msg_size() / sizeof(T);
 
-          std::unique_ptr<T[]> buf0 = (child0 != -1)
-                ? std::unique_ptr<T[]>(new T[std::min(nelem_per_maxmsg,nelem)])
-                : nullptr;
-          std::unique_ptr<T[]> buf1 = (child1 != -1)
-                ? std::unique_ptr<T[]>(new T[std::min(nelem_per_maxmsg,nelem)])
-                : nullptr;
+          auto buf0 = (child0 != -1)
+            ? static_cast<T*>(std::malloc(sizeof(T) * std::min(nelem_per_maxmsg, nelem)))
+            : nullptr;
+          auto buf1 = (child1 != -1)
+            ? static_cast<T*>(std::malloc(sizeof(T) * std::min(nelem_per_maxmsg, nelem)))
+            : nullptr;
 
           auto reduce_impl = [&,this](T* buf, size_t nelem) {
             MADNESS_ASSERT(nelem <= nelem_per_maxmsg);
             SafeMPI::Request req0, req1;
             Tag gsum_tag = world_.mpi.unique_tag();
 
-            if (child0 != -1) req0 = world_.mpi.Irecv(buf0.get(), nelem*sizeof(T), MPI_BYTE, child0, gsum_tag);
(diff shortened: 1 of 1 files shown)


src/madness/world/worldgop.h
```diff
@@ -785,8 +785,12 @@ namespace madness {
           world_.mpi.binary_tree_info(0, parent, child0, child1);
           const std::size_t nelem_per_maxmsg = max_reducebcast_msg_size() / sizeof(T);
 
-          auto buf0 = std::unique_ptr<T[]>(new T[nelem_per_maxmsg]);
-          auto buf1 = std::unique_ptr<T[]>(new T[nelem_per_maxmsg]);
+          std::unique_ptr<T[]> buf0 = (child0 != -1)
+                ? std::unique_ptr<T[]>(new T[std::min(nelem_per_maxmsg,nelem)])
```
JonathonMisiewicz, line 789: Does `std::make_unique` not work here? The less `new`, the better.

EricaCMitchell: I'll test timing with `make_unique`, when I ran tests with it by just switching the original code it resulted in longer test times in MPQC.

JonathonMisiewicz: That's annoying. For future reference, if there is a slowdown, there's a good chance [this](https://stackoverflow.com/questions/49818536/stdmake-unique-causes-big-slowdown) is the cause. Not that we can do anything about that as long as we maintain C++17 support.

EricaCMitchell: Haven't seen a difference in the timings between the two in the MPQC test cases. I'll push the `std::make_unique` implementation.

evaleev: I would avoid `std::make_unique` due to the issues discussed in Jonathon's link ... no reason to pay for value initialization of primitive types.

In fact, since this function only makes sense when `T` has trivial default ctor and trivial copy ctor (otherwise just copying bytes from MPI will not produce valid `T` objects) then we should switch to aligned malloc ... This will avoid us having to pay for initialization in the case of T=std::complex.

EricaCMitchell: This should now be accomplished with `std::malloc`. I tried `std::aligned_alloc` but the alignment constraint using `alignof(T)` resulted in a `nullptr` return when child process was present.

evaleev: > I tried `std::aligned_alloc` but the alignment constraint using `alignof(T)` resulted in a `nullptr` return when child process was present.

possibly due to asking for size that is not a multiple of alignment? See: https://en.cppreference.com/w/cpp/memory/c/aligned_alloc#Notes

EricaCMitchell: Double-checked, that doesn't seem to be the issue. From the code below and running h2-sci-x2c-1root-np2.json
```
std::cout << "Alignment: " << alignof(T) << " \n"
                << "Test: " << (sizeof(T) * std::min(nelem_per_maxmsg, nelem)) % alignof(T) << std::endl;
```
the output shows that the size is always a multiple of alignment.

evaleev: the problem with the current solution is that alignment of `T` is not guaranteed ... I pushed changes to reduce and concat0 that hopefully address it, but not tested, please test, report

EricaCMitchell: Using the most recent changes, the test set of *sci-x2c* in MPQC is decreased from 27.2s to 13.9s compared to the unaligned malloc.

evaleev (approved):

PR #552, opened 2024-10-11
replace of the "fence" changes from PR #550

evaleev:
#550 (namely, https://github.com/m-a-d-n-e-s-s/madness/commit/6ab44846c4e2e603d24e731c5bd7f18cbf22fbaf) introduced change in fence semantics that ensure that:
- local # of tasks is zero after reading AM counts
- repeat global termdet after epilogue/cleanup

this caused hangs in various parts of TiledArray when running in Github Action Runners (but not elsewhere). https://github.com/m-a-d-n-e-s-s/madness/commit/93a9a5cec2a8fa87fba3afe8056607e6062a9058 reverted it. This PR contains these changes for reconsideration.

### DIFF
--- a/src/madness/world/worldgop.cc
+++ b/src/madness/world/worldgop.cc
@@ -47,12 +47,13 @@ namespace madness {
     /// constant over two traversals.  We are then we are sure
     /// that all tasks and AM are processed and there no AM in
     /// flight.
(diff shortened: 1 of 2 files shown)


devreal:
@evaleev Is this still relevant or can we close the PR?

Issue #558, opened 2024-11-17
`aligned_alloc` may not be supported: `error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?`


barracuda156:
```
In file included from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/world/MADworld.h:46,
                 from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/tensor/elem.h:36,
                 from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/tensor/tensor_lapack.h:212,
                 from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/tensor/lapack.cc:55:
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/world/worldgop.h: In member function 'std::vector<T> madness::WorldGopInterface::concat0(const std::vector<T>&, size_t)':
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/world/worldgop.h:971:38: error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?
  971 |                                 std::aligned_alloc(sizeof(void *), bufsz)),
      |                                      ^~~~~~~~~~~~~
      |                                      aligned_union
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_science_madness/madness/work/madness-67ea7b5b533583ba7e0989cff91df738fda8b85b/src/madness/world/worldgop.h:974:38: error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?
  974 |                                 std::aligned_alloc(sizeof(void *), bufsz)),
      |                                      ^~~~~~~~~~~~~
      |                                      aligned_union
[...]

PR #559, opened 2024-11-18
Fix for case where no aligned_alloc

EricaCMitchell:
Addresses issue #558.

If HAVE_POSIX_MEMALIGN then use `posix_memalign` over `aligned_alloc`.


evaleev (approved):
@EricaCMitchell strictly speaking, this is backwards:
- we want to prefer standard things to non-standard things, and we want to prefer C++ to C.
- so std::aligned should be used first. But this means we should check for it in cmake
- cmake provides no good 1-liner to check for C++ functions reliably so need O(10) lines.

This will work. The notes are for the future. Be OCD.

PR #561, opened 2024-12-02
Add matrix_dot function

EricaCMitchell:
### Description
Adds `matrix_dot` to perform an inner product without conjugation in the bra.

### Details
- [X] Introduces `matrix_dot` functions in vmra.h
- [X] Copies `inner_local` in funcimpl.h to `dot_local` removing complex conjugation in `dot_local_X`
- [X] Adds checks using `MADNESS_CHECK` to testvmra.cc so that test fails if error threshold is not met
- [X] Update some structs to be explicitly defaulted

### DIFF
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -164,8 +164,9 @@ namespace madness {
 	        _coeffs(coeff), _norm_tree(norm_tree), _has_children(has_children), dnorm(dnorm), snorm(snorm) {
         }
 
-        FunctionNode(const FunctionNode<T, NDIM>& other) {
-            *this = other;
+        FunctionNode(const FunctionNode<T, NDIM>& other) :
+            _coeffs(other._coeffs), _norm_tree(other._norm_tree), _has_children(other._has_children),
+            dnorm(other.dnorm), snorm(other.snorm) {
         }
 
         FunctionNode<T, NDIM>&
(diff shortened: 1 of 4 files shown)


evaleev (changes requested):
@EricaCMitchell sorry, no go.
- MADNESS lacks standard format convention. So the code is pretty much formatted randomly. (some ppl even use tabs, which is makes the code look different in different editors ... maybe wrong default still better than no default???)
- Never good idea to reformat existing code. Makes code review or history tracking impossible.

Please extract the new changes out and resubmit.

EricaCMitchell:
I reverted the formatting. Although there's no standard format, I am getting a lot of clang-tidy suggestions, such as: [performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-value-param.html), [readability-avoid-const-params-in-decls](https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html), [google-explicit-constructor](https://clang.llvm.org/extra/clang-tidy/checks/google/explicit-constructor.html), and [cppcoreguidelines-pro-type-member-init](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html). Is it worth trying to fix some of these?

evaleev (approved):

PR #563, opened 2024-12-04
[jumbo] PBC

evaleev:
Major features:
- support for mixed (some open, some periodic) boundary conditions
- support for range-restricted convolutions needed to truncate the operator range to the domain of validity of finite k-point simulation
- misc cleanup
  - nuclear density function fix
    - Added `rscale`
    - Replaced the R parameter with the cell : more readable and removes the assumption that the cell is a centered cube
    - Removes the hardcoded assumption that we're in PBC.


src/madness/mra/operator.h
```diff
@@ -1061,7 +1050,7 @@ namespace madness {
                 ops[mu].setfac(coeff(mu)/c);
 
                 for (std::size_t d=0; d<NDIM; ++d) {
-                  ops[mu].setop(d,GaussianConvolution1DCache<Q>::get(k, expnt(mu)*width[d]*width[d], 0, isperiodicsum));
+                  ops[mu].setop(d,GaussianConvolution1DCache<Q>::get(k, expnt(mu)*width[d]*width[d], 0, lattice_sum[d]));
```
evaleev, line 1053: @robertjharrison while trying to deal with FunctionImpl::do_apply PBC issues I ended up tackling the mixed BC handling in SeparatedConvolution ... is there more to it than just toggling lattice sum flag for each `setop`? A pair of eyes on this commit would be good. @JonathonMisiewicz u2

robertjharrison:
Huh.

Prolly also how the operator displacements are generated in func defaults
(?).   Also if the aspect ratio is far from cubic (e.g., finite slab width
is many multiples of the periodic width) then the sorting of displacements
and the screening algorithm in do apply could be optimized (but I think
this is just performance).

I can read thru this later in the week.

evaleev:
> Prolly also how the operator displacements are generated in func defaults (?).

I went through that code, I think it's fine. I presume we only want to generate displacements for 2 cases only: all non-periodic and all-periodic, with mixed cases handled by the periodic displacements (optimizing for mixed case would just reduce the number of replacements). Or, do we want to generate the displacements for the specific boundary conditions FunctionDefaults was initialized for (i.e. assume the boundary conditions are immutable?).

I think the spherical shell-based screening logic should be OK in `do_apply` in either case. We won't be as efficient as could be but OK, let's chase correctness first.

> Also if the aspect ratio is far from cubic (e.g., finite slab width is many multiples of the periodic width) then the sorting of displacements and the screening algorithm in do apply could be optimized (but I think this is just performance). I can read thru this later in the week.

I also touched that code, and should be OK.

fbischoff:
@Ed: for the 6d convolution I implemented a function do_apply_directed_screening(), as the result of op*f is not isotropic, even if op is. Maybe that comes in handy in your case too.




------------------------------------------------------
Florian Bischoff
Humboldt-Universitaet zu Berlin, Institut fuer Chemie
Unter den Linden 6, 10099 Berlin
phone: +49-30-2093-82708
office: Brook-Taylor-Str 2, room 3'322
***@***.***
https://tinyurl.com/bischofflab

evaleev:
> @ed: for the 6d convolution I implemented a function do_apply_directed_screening(), as the result of op*f is not isotropic, even if op is. Maybe that comes in handy in your case too.

@fbischoff what is the difference with "regular" `do_apply()`? Even for standard `do_apply()` `op*f` is not isotropic ... there was a bug in the shell screening logic of `do_apply()` that I fixed at some point, maybe that was the issue that motivated the introduction of `do_apply_directed_screening()`??

fbischoff:
@evaleev the directed screening is tighter as it might truncate the expansion in one direction earlier than in the other: it won’t wait for all boxes in the shell to drop below the threshold. See my notes from that time. It made some difference for the 6d case.

￼


------------------------------------------------------
Florian Bischoff
Humboldt-Universitaet zu Berlin, Institut fuer Chemie
Unter den Linden 6, 10099 Berlin
phone: +49-30-2093-82708
office: Brook-Taylor-Str 2, room 3'322
***@***.***
https://tinyurl.com/bischofflab

src/madness/external/muParser/muParserTokenReader.cpp
```diff
@@ -807,7 +807,7 @@ namespace mu
 		m_pParser->m_vStringBuf.push_back(strTok); // Store string in internal buffer
     a_Tok.SetString(strTok, m_pParser->m_vStringBuf.size());
 
-    m_iPos += (int)strTok.length() + 2 + (int)iSkip;  // +2 wg Anf�hrungszeichen; +iSkip f�r entfernte escape zeichen
+    m_iPos += (int)strTok.length() + 2 + (int)iSkip;  // +2 wg Anf�hrungszeichen; +iSkip f�r entfernte escape zeichen
```
JonathonMisiewicz, line 810: GH doesn't render the umlaut-u doesn't on my Mac.

src/madness/mra/bc.h
```diff
@@ -0,0 +1,188 @@
...
+template <std::size_t NDIM> class BoundaryConditions {
+private:
+  // Used to use STL vector but static data on  a MAC was
+  // causing problems.
+  BCType bc[NDIM * 2];
+
+public:
+  /// Constructor. Default boundary condition set to free space
+  BoundaryConditions(BCType code = BC_FREE) {
+    for (std::size_t i = 0; i < NDIM * 2; ++i)
```
JonathonMisiewicz, line 79: Why not use `std::fill` or `std::fill_n` here? Same question for the assignment operator. Not sure if I'm missing something.

src/madness/mra/bc.h
```diff
@@ -0,0 +1,188 @@
...
+  static const char *code_as_string(BCType code) {
+    static const char *codes[] = {"zero",      "periodic",     "free",
+                                  "Dirichlet", "zero Neumann", "Neumann"};
+    return codes[code];
+  }
+
+  /// Convenience for application of integral operators
+
+  /// @return Returns a vector indicating if dimensions [0, ND) are periodic
+  template <std::size_t ND = NDIM>
```
JonathonMisiewicz, line 130: Is there a use case for ND != NDIM?

src/madness/mra/bc.h
```diff
@@ -0,0 +1,188 @@
...
+    if (d == NDIM - 1)
+      s << ")";
+    else
+      s << ", ";
+  }
+  return s;
+}
+
+template <std::size_t NDIM>
+std::array<bool, NDIM> no_lattice_sum() {
```
JonathonMisiewicz, line 180: Flagging this method as one that's likely to need edits.

src/madness/mra/leafop.h
```diff
@@ -160,21 +160,27 @@ struct ElectronCuspyBox_op : public Specialbox_op<T, NDIM> {
...
+          Key<NDIM / 2> key2;
+          key.break_apart(key1, key2);
+          int ll = this->get_half_of_special_level();
+          if (ll < f->get_initial_level())
+            ll = f->get_initial_level();
+          if (key.level() > ll) {
+            if (key1 == key2)
+              return true;
+            else
+              return false;
```
JonathonMisiewicz, line 177:
```suggestion
            return (key1 == key2);
```

src/madness/mra/leafop.h
```diff
@@ -160,21 +160,27 @@ struct ElectronCuspyBox_op : public Specialbox_op<T, NDIM> {
...
+          } else {
+            if (key1.is_neighbor_of(key2, bperiodic))
+              return true;
+            else
+              return false;
```
JonathonMisiewicz, line 182:
```suggestion
            return (key1.is_neighbor_of(key2, bperiodic));
```

src/madness/mra/gfit.h
```diff
@@ -522,7 +522,7 @@ class GFit {
 		// in the moment list
 		//
 		// <r^i|gj> cj = <r^i|exact-remainder>
-		const long nmom = 1;
```
JonathonMisiewicz, line 525: Please edit the comment above. We absolutely do not recommend NMOM=1.

src/madness/mra/key.h
```diff
@@ -174,6 +176,32 @@ namespace madness {
             return dist;
         }
 
+        uint64_t
```
JonathonMisiewicz, line 179: I think this code is using the BC to determine how to interpret boxes outside the simulation cell: in the periodic case, shift them one closer to the unit cell if outside it. This would seem to assume that you're at most within 1 simulation cell in the periodic case, and 0 otherwise.

If so, please add comments. I would never have understood this if we hadn't talked about this a few hours ago.

src/madness/mra/mraimpl.h
```diff
@@ -3233,8 +3233,13 @@ template <typename T, std::size_t NDIM>
         return true;
     }
 
+    static inline bool enforce_in_volume(Level n, Translation& l) {
```
JonathonMisiewicz, line 3236: Maybe rename to `in_unit_cell`?

JonathonMisiewicz (commented):
Reviewed most of the code, but it sounded like the new technology in `operator.h` is not ready for review, so I only gave that a casual check.

src/madness/mra/funcimpl.h
```diff
@@ -4771,84 +4781,185 @@ template<size_t NDIM>
         void do_apply(const opT* op, const keyT& key, const Tensor<R>& c) {
```
evaleev, line 4781: @robertjharrison much of #563 does not need your eyes, but this (=the whole enchilada) does .. might be more efficiento to walk you through

evaleev:
approved offline by @robertjharrison

PR #567, opened 2025-01-16
muopxv_fast way inaccurate when used with operator blocks with medium-magnitude blocks, due to excessively-loose approximation of T

### LINKED ISSUE #566: NS operator application uses overly loose threshold for T operator singular values
`muopxv_fast` uses `tol/(Rnorm*NDIM)` as threshold for SVD eigenvalues of R, but for T it uses `tol/(Rnorm*Tnorm*NDIM^2)`. Prior to https://github.com/m-a-d-n-e-s-s/madness/commit/3d2555d0ed01766f6d16b393b3e341dc8b32244c the same threshold, `tol/(Rnorm*NDIM)`, was used for both R and T.

This introduces severe artifacts in application of range-restricted kernels where operator norms are decay slowly with `n` at the edge of the range and, when applied to fine-scale boxes accurate application at each of the many intermediate scales is important.

evaleev:
resolves #566

### DIFF
--- a/src/madness/mra/operator.h
+++ b/src/madness/mra/operator.h
@@ -432,7 +432,7 @@ namespace madness {
                          const Tensor<T>& f, const Tensor<T>& f0,
                          Tensor<TENSOR_RESULT_TYPE(T,Q)>& result,
                          Tensor<TENSOR_RESULT_TYPE(T,Q)>& result0,
-                         double tol,
+                         const double tol,
                          const Q mufac,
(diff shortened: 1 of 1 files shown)


evaleev:
@fbischoff please have a look at this ... the original change that led to this was a long time ago so I'm sure it's hard to recall the reasons ... but the change was not noticed which makes me wonder whether this is something special about most common operators (BSH, Poisson) that makes the bug not noticeable and suggests that there is some structure that is still not being exploited. My change has potential to make these common scenarios potentially slightly slower. cc: @robertjharrison

PR #568, opened 2025-01-16
[ci] bump to ubuntu 24.04 and gcc 13

evaleev:
gcc 9 is insanely old


evaleev:
will be able to use gcc 14 only when new RapidJSON release occurs: https://github.com/Tencent/rapidjson/issues/2277

PR #571, opened 2025-01-23
Fix to do_dot_localX for GenTensor

EricaCMitchell:
### Description
GenTensor does not do [trace of <this|rhs>](https://github.com/m-a-d-n-e-s-s/madness/blob/577c2980cbc83ab49ab9eb99d69a3bbdfd09fa0f/src/madness/tensor/lowranktensor.h#L510) for complex LowRankTensor.

### Details
- [X] Add `std::enable_if_t` to [`do_dot_localX`](https://github.com/m-a-d-n-e-s-s/madness/blob/577c2980cbc83ab49ab9eb99d69a3bbdfd09fa0f/src/madness/mra/funcimpl.h#L5716) for GenTensor implementation.
- [X] Replace `trace` with `trace_conj` in [`do_dot_localX`](https://github.com/m-a-d-n-e-s-s/madness/blob/577c2980cbc83ab49ab9eb99d69a3bbdfd09fa0f/src/madness/mra/funcimpl.h#L5743) for GenTensor implementation. `trace` does not exist for GenTensor.
- [X] Disables HAVE_GENTENSOR implementation, implementation uses GenTensor full_tensor() which will throw if GenTensor does not have Tensor.

### DIFF
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -50,7 +50,6 @@
 #include <madness/mra/function_factory.h>
 
 #include "leafop.h"
-#include "type_data.h"
 
 namespace madness {
     template <typename T, std::size_t NDIM>
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -50,6 +50,7 @@
 #include <madness/mra/function_factory.h>
 
 #include "leafop.h"
+#include "type_data.h"
 
 namespace madness {
     template <typename T, std::size_t NDIM>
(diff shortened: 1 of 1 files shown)


fbischoff:
I've submitted a suggestion for fixing the GenTensor/Tensor issue, without having the preprocessor directives.

coeffT is always a GenTensor, which is an alias to Tensor if ENABLE_GENTENSOR is disabled. It has a number of no-op functions for compatibility like full_tensor() which turns a GenTensor into a Tensor. 

If ENABLE_GENTENSOR is enabled, the GenTensor can hold either a Tensor, an SVDTensor or a TensorTrain, determined at runtime. full_tensor() will return a Tensor if GenTensor holds a Tensor, and will throw if GenTensor holds a different type of object, like SVDTensor

These are the possible cases:
 - ENABLE_GENTENSOR=false: GenTensor is derived from Tensor and full_tensor() returns itself.
 - ENABLE_GENTENSOR=true: GenTensor holds a Tensor, full_tensor() returns a Tensor
 - ENABLE_GENTENSOR=true: GenTensor holds something else, full_tensor() throws.

fbischoff (approved):
If it works for you feel free to merge

PR #580, opened 2025-02-14
Atomic SAP Potentials, Fix Interpolator Bug

JonathonMisiewicz:
1. Adds SAP interpolation technology. All atomic SAP interpolation tables are now created... **Is it worth changing this to only create when necessary?**
2. Fixes a bug in interpolation 1d caused by me thinking sort function arguments where _indices_ rather than _values_.

I've also had to add a copyright notice to use @susilehtola's SAP tables. There are alternative tables in the SI that, AFAIK, aren't copyright restricted, but those cover fewer atoms. Please confirm this is acceptable.


JonathonMisiewicz:
I'm finding some anomalies that I swear weren't there last night. Let's hold off on merging until I have this locked down... Reviews are still welcome.

JonathonMisiewicz:
We're good to proceed as normal now.

src/madness/chem/potentialmanager.cc
```diff
@@ -135,8 +136,41 @@ double WignerSeitzPotentialFunctor::operator()(const coord_3d &r) const {
...
+    const Atom &atom,
+    const BoundaryConditions<3> &bc,
+    const Tensor<double> &cell,
+    int special_level)
+    : atom(atom), bc_(bc), cell(cell), special_level_(special_level),
+      special_points_({atom.get_coords()}) {
+}
+
+double SAPFunctor::operator()(const coord_3d &x) const {
+  const double zero_cutoff = 1e-10;
```
susilehtola, line 152: I am not sure why you need this

susilehtola (approved):
Looks reasonable

src/madness/chem/potentialmanager.cc
JonathonMisiewicz: I could well be not understanding some numerical analysis here, but if `r = 0`, l. 168 may try to compute `0 / 0`, no?

susilehtola: Yes, for sure; yet, I thought that the multiresolution scheme does not put any points at the nucleus. Still, I think `1e-10` is a hardcoded cutoff that is somewhat meaningless; it could also be `DBL_EPSILON` or a smaller number since its reciprocal is still a normal number.

evaleev: @susilehtola when projecting to MW basis this functor will be called for points whose coordinates are controlled by the Gauss-Legendre quadrature so unlikely that you will hit the nucleus exactly, but must consider this as a possibility

src/madness/chem/potentialmanager.cc
```diff
@@ -135,8 +136,41 @@ double WignerSeitzPotentialFunctor::operator()(const coord_3d &r) const {
...
+    auto maxY = bc_.is_periodic()[1] ? static_cast<int>(ceil(interpolator.get_hi() / ext_y)) : 0;
+    auto maxZ = bc_.is_periodic()[2] ? static_cast<int>(ceil(interpolator.get_hi() / ext_z)) : 0;
+    for (int dx = -maxX; dx <= maxX; dx++) {
+      double xx = atom.x + dx*ext_x - x[0];
+      for (int dy = -maxY; dy <= maxY; dy++) {
+        double yy = atom.y + dy*ext_y - x[1];
+        for (int dz = -maxZ; dz <= maxZ; dz++) {
+          double zz = atom.z + dz*ext_z - x[2];
+          auto r = std::sqrt(xx * xx + yy * yy + zz * zz);
+          if (r > zero_cutoff and r < interpolator.get_hi()) sum += interpolator(r) / r;
```
evaleev, line 168: @JonathonMisiewicz the proper way to avoid the need for `zero_cutoff` is to use smoothed_potential here ...

src/madness/misc/interpolation_1d.h
```diff
@@ -247,6 +253,8 @@ template <typename T> class CubicInterpolationTable {
       return (a[i + 1] + y1 * a[i + 2]) + yy * (a[i + 3] + y1 * a[i + 4]);
   }
 
+  double get_hi() const { return hi; }
```
evaleev, line 256: while at it add `get_lo()`

src/madness/chem/potentialmanager.h
```diff
@@ -202,6 +203,28 @@ class WignerSeitzPotentialFunctor : public FunctionFunctorInterface<double,3> {
...
+  int special_level_ = 15;
+ public:
+  /// Generic constructor, can handle open and periodic boundaries
+  /// \param molecule atoms
+  /// \param bc boundary conditions
+  /// \param cell simulation cell (unit cell, if periodic)
+  /// \param special_level the initial refinement level
+  /// \param rscale setting `rscale>1` will make a nucleus larger by a factor of \p rscale (in other words, `rcut` is multiplied by the inverse of by this)
+  SAPFunctor(const Atom& atom,
+                        const BoundaryConditions<3>& bc = FunctionDefaults<3>::get_bc(),
```
evaleev, line 221: formatting ...

src/madness/chem/CMakeLists.txt
```diff
@@ -90,6 +91,7 @@ set(MADCHEM_SOURCES
     pointgroupsymmetry.cc
     potentialmanager.cc
     polynomial.cc
+    SAP_interpolators.cc
```
evaleev, line 94: perhaps trim down to `SAP.*`

evaleev (changes requested):
see embedded comments

src/madness/chem/potentialmanager.cc
susilehtola: Yes but like you point out below, MADNESS doesn't use the Coulomb potential but the smoothed potential that doesn't have this issue, so the cutoff can be removed altogether...

evaleev (approved):

PR #584, opened 2025-02-28
Exchange for large molecules

hborchert:
Adjusted multiworld algorithm for exchange that computes exchange contribution for a fixed orbital as a single task instead of tiling the exchange matrix and other changes including: tightened threshold for the coulomb operator, default lo parameter, tiling of Vpsi, applyBSH and transform amo in SCF.cc to reduce memory requirement for large molecules


dif
evaleev, line 1: presumably added by mistake?

PR #605, opened 2025-04-18
Update DFdriver

EricaCMitchell:
### Description
Updates the Dirac-Fock code to correctly read in orbitals from moldft to compute a closed-shell system. PR is purposely more minimal to keep implementation independent of MPQC.

Note, I did have working version of DFdriver using small memory exchange algorithm, but it was 10x slower than current algo due to need for better calls to vmra.h and mra.h. Seems better to leave exchange as is to have independent implementation from MPQC, which does use smallmem exchange algorithm for bispinors.

### Details
- [X] Adds point charge nucleus model, option is `nucleus == 2`
- [X] Properly computes closed-shell computation, without Kramers symmetry
- [X] Converge either due to BSH residuals or energy, density, and BSH residuals
- [X] Updates for speed of light and bohr radius in fm to CODATA 2022 values

### DIFF
--- a/src/apps/dirac/DF.cc
+++ b/src/apps/dirac/DF.cc
@@ -333,6 +333,66 @@ double myr(const coord_3d& r){
      return std::sqrt(r[0]*r[0]+r[1]*r[1]+r[2]*r[2]);
 }
 
+//Creates the (Gaussian) nuclear potential from the molecule object
+void DF::make_gaussian_potential(World& world, real_function_3d& potential){
+     if(world.rank()==0) print("\n***Making a Gaussian Potential***");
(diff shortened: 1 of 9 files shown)


evaleev (changes requested):
please make the speed of light the input parameter

robertjharrison:
I merged master into this branch, fixed issues due to code divergence since the PR was generated, tested on He and water, added sparse mul operations that Erica previously identified as key missing performance issue, retested, and submitted it all as a new PR as a clean commit on top of master in this repo.  So I am closing this PR since all of the content is now in the new one (which I will likely just immediately merge since all of the changes are in the dirac folder).

PR #633, opened 2025-10-27
fixup Function::swap

evaleev:
`Function::swap` used nonexistent `FunctionImpl::swap`. Since `Function` has reference semantics `Function::swap` should have same semantics (i.e. shallow swap). Use std::shared_ptr::swap.


evaleev:
> weird that it was not doing this from the start!

the only explanation I see is that this is not instantiated in any of the tests done by CI ... MPQC would not compile, but the compiler would not point out where the instantiation occurred, only in which header the problem occurred. Did not chase this further.

PR #658, opened 2026-03-04
Fixes found necessary for 1D PBC

### LINKED ISSUE #625: Monotonic Convolution Assumption Fails
A previous commit [removed](https://github.com/m-a-d-n-e-s-s/madness/commit/8523617d827b6278f6f01a546818da8703027601) the assumption that integral convolution was always with isotropic kernels, weakening it to the assumption that kernels are merely "monotonic" for significant `r`. This assumption is incorrect for perfectly physical potentials.

The technical definition of monotonically decreasing is important here. For purposes of the [screening procedure](https://github.com/m-a-d-n-e-s-s/madness/blob/master/src/madness/mra/funcimpl.h#L4958), monotonically decreasing means that the maximum contribution of displacements with Euclidean-displacement-norm n decreases as a function from n, e.g., the contribution from (2, 2, 2) with n = sqrt(8) must be larger than the contribution from (0, 0, 3) and (0, 1, 3) with n = sqrt(9) and sqrt(10), respectively. This is quite reasonable for cubic boxes, where the map from displacement number along an axis to physical displacement along an axis is the same for each axis; a 2 in the x position is no different in terms of real-space distance than a 2 in the z position.

Unfortunately, this assumption fails for the simple case of a Gaussian in a non-cubic box. In my test case, my box is 60 x 60 x 7.72, and I apply a single Gaussian to a spheropole (a model of a hydrogen atom with Gaussian α values of 10000 for the nucleus and 6 for the electron). For this box, the _real-space boxes_ at (0, 0, 3) and (0, 1, 3) are much closer than the _real-space boxes_ at (2, 2, 2). The map from displacement number along an axis to physical displacement now differs from each axis, and the contribution from (0, 0, 3) and (0, 1, 3) are neglected because the more distant (2, 2, 2) can be neglected. Accordingly, the summation shuts off prematurely, and I see error in the sixth decimal even with ε=1*^-10

I'm not sure if there's a good way to restore this screening...

@robertjharrison @evaleev

JonathonMisiewicz:
Closes #625. Needed for periodic MRA purposes.

- [x] `static` member variables are now initialized inline, silencing warnings.
- [x] `funcimpl.h` uses a range-based for loop, resulting in much cleaner code.
- [x]  `bmax_default(3)` has been increased for precision.
- [x] Displacements are now sorted according to distance _in real-space_, not in simulation space, provided that the user calls `set_width`. The displacement screening algorithm assumes this.
- [x] `disp_periodic` is now a `std::array` rather than a C-array. 
- [ ] Note new functions in `key.h`. Some of the old functions are special cases of these (with a Tensor of all 1's), but is it worth making the old ones call these?

### DIFF
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -4991,7 +4991,7 @@ template<size_t NDIM>
               const auto real_distsq = real_distance_squared(displacement);
               const std::uint64_t lattice_distsq = real_distsq ? 0 : lattice_distance_squared(displacement);
               if (!real_last_distsq.has_value() ||
-                  real_distsq != *real_last_distsq || (*real_last_distsq == 0 && lattice_distsq != *lattice_last_distsq)) { // Moved to next shell of neighbors
+                  !nearlyEqual(real_distsq, *real_last_distsq) || (nearlyEqual(*real_last_distsq, 0) && lattice_distsq != *lattice_last_distsq)) { // Moved to next shell of neighbors
                 if (nvalid > 0 && nused == 0 && (real_distsq > 0 || lattice_distsq > 1)) {
                   // Have at least done the input box and all first
                   // nearest neighbors, and none of the last set
--- a/src/madness/misc/CMakeLists.txt
+++ b/src/madness/misc/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 set(MADMISC_HEADERS misc.h ran.h phandler.h interpolation_1d.h cfft.h info.h gnuplot.h array_of_bools.h kahan_accumulator.h)
 set(MADMISC_SOURCES
-    checksum_file.cc position_stream.cc gprofexit.cc ran.cc cfft.cc info.cc unique_filename.cc)
+    checksum_file.cc position_stream.cc gprofexit.cc ran.cc cfft.cc info.cc numerics.cc unique_filename.cc)
 # retrieve git metadata
 include(GetGitMetadata)
 vgkit_cmake_git_metadata()
--- a/src/madness/misc/misc.h
+++ b/src/madness/misc/misc.h
@@ -51,6 +51,8 @@ namespace madness {
 
     // get the memory usage of the current process in byte
     long get_memory_usage();
+
+    bool nearlyEqual(double a, double b, double epsilon = 1e-9);
 }
 
 #endif // MADNESS_MISC_MISC_H__INCLUDED
--- a/src/madness/misc/numerics.cc
+++ b/src/madness/misc/numerics.cc
@@ -1,0 +1,21 @@
+//
+// Created by Jonathon Misiewicz on 5/16/23.
+//
+
+#include <madness/misc/misc.h>
+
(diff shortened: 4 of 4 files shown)


JonathonMisiewicz:
Would like review from **both** of you before merging this one in.

src/madness/mra/key.h
```diff
@@ -188,9 +189,17 @@ namespace madness {
             return dist;
         }
 
+        // The distance of a displacement in real space.
+        uint64_t real_distsq(const madness::Tensor<double>& widths) const {
+          uint64_t dist = 0;
+          for (std::size_t d = 0; d < NDIM; ++d) {
+              dist += widths(d) * l[d] * l[d];
```
evaleev: should this not be `(widths(d) * l[d])^2`?

src/madness/mra/funcimpl.h
```diff
@@ -5047,7 +5056,7 @@ template<size_t NDIM>
             // skip surface displacements that take us outside of the domain and/or were included in regular displacements
             // N.B. for lattice-summed axes the "filter" also maps the displacement back into the simulation cell
             if (max_distsq_reached)
-              validator = BoxSurfaceDisplacementValidator<opdim>(/* is_infinite_domain= */ op->func_domain_is_periodic(), /* is_lattice_summed= */ op->lattice_summed(), range, default_distance_squared, *max_distsq_reached);
+              validator = BoxSurfaceDisplacementValidator<opdim>(/* is_infinite_domain= */ op->func_domain_is_periodic(), /* is_lattice_summed= */ op->lattice_summed(), range, default_real_distance_squared, *max_distsq_reached);
```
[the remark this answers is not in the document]
JonathonMisiewicz: Your text is correct. Your code suggestion is not.

src/madness/mra/funcimpl.h
```diff
@@ -4981,11 +4984,15 @@ template<size_t NDIM>
...
-                  dsq != *distsq) { // Moved to next shell of neighbors
-                if (nvalid > 0 && nused == 0 && dsq > 1) {
+              // Screen out shells. We assume shells are grouped into shells so that the operator decays with shell index.
+              // Shells are indexed by least distance from box to the central box.
+              // Cells touching so much as a corner of the central box are further grouped by their lattice distance.
+              // N.B. lattice-summed decaying kernel is periodic (i.e. does decay w.r.t. r), so loop over shells of displacements sorted by distances modulated by periodicity (Key::distsq_bc)
+              const auto real_distsq = real_distance_squared(displacement);
+              const std::uint64_t lattice_distsq = real_distsq ? 0 : lattice_distance_squared(displacement);
+              if (!real_last_distsq.has_value() ||
+                  real_distsq != *real_last_distsq || (*real_last_distsq == 0 && lattice_distsq != *lattice_last_distsq)) { // Moved to next shell of neighbors
```
evaleev: "hard" comparison of reals ... should use soft compare perhaps?

evaleev (approved):

PR #677, opened 2026-03-23
Add width setting for Displacements in testopdir

### LINKED ISSUE #676: testopdir failure introduced by commit 7bec899
Hi Jonathon,

I've found a failing test: testopdir, that only occurs if the "smalltest" flag is not set, which it is on github, so it wouldn't show up there, but it showed up locally on my machine.

Best,
Florian

Summary (from Claude)

  The testopdir test (in src/madness/mra/testopdir.cc) started failing after commit 7bec8999 ("Further copilot changes"). Identified via git bisect between known-good 56fa55ba  and current master.
                                                                                                                                                                                  
  Root cause                                                                                                                                                                    

  Commit 7bec899 added std::abs() to the displacement distance calculation in src/madness/mra/key.h, in both real_distsq() and real_distsq_bc():                                  
   
  // Before (working):                                                                                                                                                            
  std::max(l[d] - 1, static_cast<Translation>(0));                                                                                                                              
  // After (broken):                                                                                                                                                              
  std::max(std::abs(l[d]) - 1, static_cast<Translation>(0));
                                                                                                                                                                                  
  The operator screening code (funcimpl.h:4932) always uses real_distsq_bc(), even for non-periodic domains (where lattice_summed() is all-false). In that case the periodicity  folding is a no-op, but translation()[d] can still be negative for displacements in the negative direction.
                                                                                                                                                                                  
  For a negative displacement like la = -2:                                                                                                                                       
  - Old: max(-2 - 1, 0) = 0 → treated as touching → included in operator sum
  - New: max(abs(-2) - 1, 0) = 1 → nonzero distance → potentially screened out                                                                                                    
                                                                              
  While abs() is mathematically correct for measuring box-to-box distance, the more aggressive screening causes the operator to exclude contributions it needs for accuracy,  producing numerical errors that exceed the reference thresholds in the test. The test uses a non-cubic cell (widths 40, 50, 80) which amplifies direction-dependent screening  differences.                                                                                                                                                                    
                                                                                                                                                                                  
  The same commit also changed max_distsq_reached and DistanceSquaredFunc types from Translation to double in displacements.h, which may interact with the screening comparisons. 
   
  Reproducing                                                                                                                                                                     
                  
  cd <build_dir>
  cmake --build . --target testopdir
  ctest -R testopdir --output-on-failure                                                                                                                                          
   
  Suggested fix                                                                                                                                                                   
                  
  Revert the key.h changes from commit 7bec899, or adjust the screening threshold to account for the corrected distance calculation.

JonathonMisiewicz:
This PR closes #676. The new screening algorithm requires (at least for non-cubic cells) that the displacements receive a copy of the cell width.

I understand @fbischoff is changing where the cell width data lives, and it's worth asking if `FunctionDefaults<NDIM>::set_cell` should update the width in `Displacements<NDIM>`. My intuition is to say yes, but I don't know how his changes affect that.

### DIFF
--- a/src/madness/mra/testopdir.cc
+++ b/src/madness/mra/testopdir.cc
@@ -153,6 +153,7 @@ int test_opdir(World& world) {
     cell(2,0)=-40; cell(2,1)=40;
     //FunctionDefaults<3>::set_cubic_cell(-20,20);
     FunctionDefaults<3>::set_cell(cell);
+	Displacements<3>::set_width(FunctionDefaults<3>::get_cell_width());
     FunctionDefaults<3>::set_k(8);
     FunctionDefaults<3>::set_thresh(1e-6);
     //FunctionDefaults<3>::set_k(10);
@@ -265,6 +266,7 @@ int testgradG(World& world) {
     const double thresh = 1e-8;
 
     FunctionDefaults<3>::set_cell(cell);
+	Displacements<3>::set_width(FunctionDefaults<3>::get_cell_width());
     FunctionDefaults<3>::set_k(k);
     FunctionDefaults<3>::set_thresh(thresh);
     FunctionDefaults<3>::set_initial_level(3);


fbischoff:
If you change the cell all functions will not be normalized anymore, and the nuclear potential will not be consistent with the molecule. I guess you want to change the cell while optimizing?

I've just started a PR where I keep the cell info in FunctionImpl, similar to k and thresh. So if you change the FunctionDefault::cell those will be inconsistent too. I was not expecting the cell to change within a single calculation.. So do I have to rethink my design?

PR #690, opened 2026-04-23
rewrite `AGENTS.md` as the canonical agent guidance

evaleev:
also: symlink CLAUDE.md, drop GEMINI.md

almost all shops now standardize around AGENTS.md . The current iteration avoids restating things that are easily greppable, and only states things that are hard(er) for agent introspect and/or corrects the incorrect defaults. The more goes here the higher the chance the content becomes stale/out-of-sync


fbischoff (approved):
suggestions: 
 - other codes that rely on MADNESS: tequila, frayedends
 - make sure the fence in blocks like if (world.rank()==0) {..}
 - other common build directories: ../compile/madness-{debug,options}

PR #694, opened 2026-05-13
Provide a wrapper for atomic shared pointers

devreal:
C++20 has deprecated free-standing atomic operations on `std::shared_ptr` and instead introduced `std::atomic<std::shared_ptr>`. Introduce a wrapper `atomic_shared_ptr` that abstracts away the handling of different versions of C++.

This PR does not change the required C++ version but is needed when MADNESS is pulled into a project that requires new C++.

### DIFF
--- a/src/madness/world/CMakeLists.txt
+++ b/src/madness/world/CMakeLists.txt
@@ -11,7 +11,7 @@ set(MADWORLD_HEADERS
     uniqueid.h worldprofile.h timers.h binary_fstream_archive.h mpi_archive.h 
     text_fstream_archive.h worlddc.h mem_func_wrapper.h taskfn.h group.h 
     dist_cache.h distributed_id.h type_traits.h function_traits.h stubmpi.h 
-    bgq_atomics.h binsorter.h parsec.h meta.h worldinit.h thread_info.h
+    bgq_atomics.h binsorter.h parsec.h meta.h atomic_shared_ptr.h worldinit.h thread_info.h
     cloud.h test_utilities.h timing_utilities.h units.h ranks_and_hosts.h)
 set(MADWORLD_SOURCES
     madness_exception.cc world.cc timers.cc future.cc redirectio.cc
--- a/src/madness/world/atomic_shared_ptr.h
+++ b/src/madness/world/atomic_shared_ptr.h
@@ -23,14 +23,16 @@ namespace madness {
 /// \c use_count, equality comparison) plus a single extension, \c exchange(),
 /// needed for the move-optimisation in \c Future::get()&&.
 ///
-/// \note User-declared copy constructor and copy-assignment operator suppress
-/// the implicit move constructor and move-assignment operator (and
-/// \c std::atomic is itself non-movable in the C++20 backend), so move
-/// operations silently fall back to copy.
+/// \note Move operations are explicitly provided: they atomically take the
+/// value from the source (leaving it empty) via a relaxed-order \c exchange.
+/// This is necessary because (a) user-declared copy operations would otherwise
+/// suppress the implicit move operations, and (b) \c std::atomic is non-movable
+/// in the C++20 backend, so the implicit move would silently fall back to copy.
 ///
-/// \note All loads are \c memory_order_relaxed, relying on the caller to provide necessary
-/// ordering guarantees. The exception is \c exchange(), which defaults to \c memory_order_seq_cst to match the pre-C++20
-/// \c std::atomic_exchange behaviour it replaces.
+/// \note All loads/stores and move operations use \c memory_order_relaxed,
+/// relying on the caller to provide necessary ordering guarantees. The exception
+/// is \c exchange(), which defaults to \c memory_order_seq_cst to match the
+/// pre-C++20 \c std::atomic_exchange behaviour it replaces.
 template<typename T>
 class atomic_shared_ptr {
 #if defined(__cpp_lib_atomic_shared_ptr)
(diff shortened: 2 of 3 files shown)


evaleev (commented):
## Overview

`std::atomic_exchange(std::shared_ptr<T>*, …)` is deprecated in C++20 in favour of `std::atomic<std::shared_ptr<T>>`. This PR introduces `madness::atomic_shared_ptr<T>` in a new header `src/madness/world/atomic_shared_ptr.h`, switches `Future<T>::f` from `std::shared_ptr<FutureImpl<T>>` to the wrapper, and replaces the deprecated call site in `Future::get() &&`. Backend chooses `std::atomic<std::shared_ptr<T>>` when `__cpp_lib_atomic_shared_ptr` is available, otherwise the deprecated free functions (with localized pragma suppression). The motivating use case — staying compatible with downstream consumers that build at C++20 with `-Werror` — is legitimate and was foreshadowed by the FIXME at the call site.

## Correctness

- **All reads of `Future::f` are now atomic.** Previously only the one `atomic_exchange` site was atomic; reads such as `if (f)`, `f->probe()`, `f.use_count()` were plain `shared_ptr` accesses racing against the atomic exchange in `get() &&` — technically UB. The wrapper closes that hole; nice side-benefit beyond the warning fix.
- **`reset(T* p = nullptr)` is marked `noexcept` but allocates internally.** `std::shared_ptr<T>(p)` constructs a control block and can throw `std::bad_alloc`. The original `shared_ptr::reset(T*)` is not `noexcept`. As written, an OOM here calls `std::terminate`. Suggest dropping `noexcept` from `reset`.
- **`reset()` with default `nullptr` is not equivalent to `shared_ptr::reset()`.** `shared_ptr<T>(static_cast<T*>(nullptr))` creates a control block with `use_count()==1` and a deleter; `shared_ptr<T>()` is empty (`use_count()==0`). Today nothing in the codebase calls `f.reset()` with no argument, so this is latent, but it's a behavioural divergence from the type it's mimicking. Easy fix:
  ```cpp
  void reset(T* p = nullptr) { do_store(p ? std::shared_ptr<T>(p) : std::shared_ptr<T>()); }
  ```
- **Race analysis of `Future::get() &&`** still holds: the rvalue path takes `value_ref = f->get()` (the temporary `shared_ptr` from `do_load()` keeps the impl alive through the full-expression), then `f.exchange({})` atomically removes `ptr_`'s ref; if `fcopy.use_count() == 1`, no concurrent copier observed the future, safe to move. Good.

## Style / project conventions

- **Use `MADNESS_PRAGMA_GCC` / `MADNESS_PRAGMA_CLANG`** rather than raw `#pragma`. The code displaced from `future.h` used these macros; the new header reintroduces raw pragmas — inconsistent and harder to grep for. There are now four near-identical pragma blocks; consider a small internal helper macro or, at minimum, swap to the project macros.
- The `__cplusplus >= 202002L` guard around the deprecation suppressions is correct (the free functions are only deprecated at C++20), but it's worth a one-line comment clarifying that — otherwise the next reader wonders why we suppress deprecation only at C++20 while sitting in a pre-C++20 fallback branch.
- The unrelated trailing-whitespace touch at `future.h:123` is harmless but doesn't belong in the diff.

## Performance

- Every observation of `f` now pays an atomic load + a refcount inc/dec (both fall-backs are typically lock-protected). Hot paths in `future.h` like `is_local()`, `is_default_initialized()`, `f != other.f`, and `if (f ? f->probe() : …)` each do two atomic loads now. This is the price of correctness, and matches what would have to happen if `Future::f` truly needs to be atomic — but the PR description does not flag the regression. Worth measuring on whatever existing benchmark exercises `Future` heavily (e.g. `test_future3`, task scheduling micro-bench) before merging.
- `memory_order_relaxed` on load/store with `memory_order_seq_cst` (default) on `exchange` matches the prior semantics (the old code used `std::atomic_exchange` which is seq_cst) and is defensible: identity-only access of `ptr_`; ordering of the payload (the `FutureImpl::assigned` flag) is supplied separately. A one-line comment to this effect would save future readers from second-guessing.

## Tests

- No dedicated unit test for `atomic_shared_ptr`. Coverage is implicit via `test_future*.cc`. Given the C++20-conditional code path, a tiny test that exercises copy/assign/exchange/`use_count` would be cheap insurance and would exercise both branches on different toolchains. Not a blocker.
- Has the PR been compiled and tested with `-DMADNESS_WERROR=ON` against (a) a libstdc++ recent enough to define `__cpp_lib_atomic_shared_ptr` and (b) one without? The whole point of the change is to clear that build, so explicit confirmation in the PR description would be welcome.

## Suggested follow-ups (not blockers)

- Move semantics: copy ctor/assign are user-declared, so implicit move ctor/assign are deleted (and the C++20 backend can't define move ctor anyway — `std::atomic` isn't movable). Move-construction will silently fall back to copy. Acceptable, but worth one sentence in the doc string.
- Doxygen sentence on `operator->()` — "Safe because the caller's `Future` keeps the underlying object alive" — slightly misleads. The real safety argument is that `do_load()` returns a `shared_ptr` temporary whose lifetime extends to the end of the full expression, holding a refcount that keeps the impl alive across the dereference. Tightening this comment would help.

## Verdict

Direction is right; the wrapper is appropriately minimal. Address the `noexcept`/`reset()` correctness items and switch to the existing `MADNESS_PRAGMA_*` macros for project consistency, and this is ready to merge.

evaleev:
rebased onto master + added move ctor/assignment for `atomic_shared_ptr`

evaleev (approved):

PR #695, opened 2026-05-21
WorldGopInterface::is_in_do_cleanup() — let deferred destructors skip cross-rank handshakes

evaleev:
## Summary

Add a per-`WorldGopInterface` bool flag, set via an RAII scope in `fence_impl` around `deferred_->do_cleanup()`, and a public accessor `is_in_do_cleanup()`. Destructors invoked from within the deferred-cleanup phase can query the flag and skip cross-rank collective synchronization that the fence has already established.

## Why

`fence_impl` runs the global-termination protocol (drain loop + Dykstra-style nsent/nrecv tree handshake), so by the time `do_cleanup` runs, all ranks are at the same point with no AM in flight. `defer_deleter_to_next_fence` is, by contract, used collectively, so every rank's deferred list holds the same set of objects at this point and the matching cleanup actions on each rank are issued in lockstep.

Under those two invariants, a destructor running here may delete its object directly rather than route through a collective handshake like `lazy_sync` — the handshake's purpose (avoiding deletion while a peer may still send AM addressed to the object) is already satisfied by the fence.

Downstream consumer ([ValeevGroup/tiledarray#550](https://github.com/ValeevGroup/tiledarray/pull/550)) uses this to make `DistArray::lazy_deleter` bypass `lazy_sync` in the deferred-cleanup path. That eliminates a class of bugs where `lazy_sync_children` tasks survive the fence (they're scheduled after the drain loop) and are later run against freed state when the world is destroyed before peers reach a matching fence — most visibly when TA-`einsum` per-Hadamard sub-Worlds are torn down at function exit or during exception unwind, tripping `~WorldObject`'s `World::exists(&world)` assertion.

## What

- `worldgop.h`: private `bool in_do_cleanup_ = false`, public `is_in_do_cleanup()`.
- `worldgop.cc`: tiny RAII scope in `fence_impl` that flips the flag around `deferred_->do_cleanup()`, exception-safe.

## Test plan

- [x] Local downstream MPQC repro (bimal.json CSV-CCk traced expression with arena-ToT operands): previously aborted in `~WorldObject` during einsum stack unwind; runs end-to-end with exit 0 after this + TA#550.
- [ ] CI green (single-rank and multi-rank).

## Note on a previous attempt

An earlier version of this PR added a `world_.taskq.fence()` after `do_cleanup()` inside `fence_impl`. That deadlocked on multi-rank because `lazy_sync` is a collective requiring matching `lazy_sync(id, ...)` on every peer for the same id — but per-rank `shared_ptr<...>` refcounts are independent and peers' deferred lists need not agree at a given fence boundary (broke TA's `expressions_complex_suite/assign_subblock_block_contract` test). The current approach removes the offending synchronization entirely rather than trying to drain it locally.

### DIFF
--- a/src/madness/world/worldgop.cc
+++ b/src/madness/world/worldgop.cc
@@ -149,7 +149,22 @@ namespace madness {
         MADNESS_ASSERT(pause_during_epilogue == false);
         epilogue();
         world_.am.free_managed_buffers(); // free up communication buffers
-        deferred_->do_cleanup();
+        // Run deferred cleanup with the in_do_cleanup_ flag set so that any
+        // destructor invoked here can detect that it runs from inside the
+        // fence's synchronization boundary and skip cross-rank handshakes
+        // (e.g. TA's DistArray::lazy_deleter checks this and bypasses
+        // lazy_sync, which would otherwise schedule lazy_sync_children tasks
+        // that the fence cannot drain and that survive into the global
+        // ThreadPool past this world's lifetime). The flag is reset on every
+        // exit path, including exceptions.
+        {
+          struct InDoCleanupScope {
+            bool& flag;
+            ~InDoCleanupScope() { flag = false; }
+          } scope{in_do_cleanup_};
+          in_do_cleanup_ = true;
+          deferred_->do_cleanup();
+        }
 #ifdef MADNESS_HAS_GOOGLE_PERF_TCMALLOC
         MallocExtension::instance()->ReleaseFreeMemory();
 //        print("clearing memory");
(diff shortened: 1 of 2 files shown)


evaleev:
Closed automatically by force-push that reset the branch to upstream master. The post-do_cleanup taskq.fence() approach forces synchronization on lazy_sync_children tasks that require a matching call on every peer rank for the same pimpl id; since per-rank shared_ptr<ArrayImpl> refcounts are independent, peer deferred lists can differ at the same fence boundary and main-world fences deadlock (observed in TA's expressions_complex_suite/assign_subblock_block_contract). The TA-einsum sub-World case that motivated this fix is now addressed in TiledArray (ValeevGroup/tiledarray#550) by double-fencing each sub-World in an RAII guard at function exit -- the ranks of a sub-World *do* reach that point in lockstep, so the cross-rank lazy_sync handshake matches up there.

PR #704, opened 2026-07-08
Modern response stack: molresponse solver + madqc driver (replaces legacy implementations)

ahurta92:
## Summary

This PR delivers the modern MADNESS response stack: the **molresponse** solver
(linear + quadratic response on the MRA protocol ladder) and the **madqc**
unified driver, replacing the three legacy response implementations.

### Commits (sliced for review clarity)
1. `refactor!:` **remove legacy molresponse implementations** — molresponse (v1)
   and molresponse_v2 (the MolresponseLib engine) are deleted; before removal,
   v2↔v3 parity was captured and validated (α agrees to ≤2e-5 rel; the full
   81-component v2 Raman tensor is archived and v3's single-component Raman
   validated against it at 0.13%).
2. `feat(molresponse):` **the solver** — protocol-ladder with honest-climb
   (budget-exhausted rungs seed finer ones; convergence labels stay honest),
   per-row property accuracy metadata, state-parallel FD via node-aligned
   subworlds (`--fd-subworlds`), an FD exchange-tensor path
   (`--fd-tensor`/`--es-tensor`, tiled), opt-in HDF5 restart I/O, and MRA
   tree/orbital visualization export (HTG / Overlapping-AMR / native coefficients).
3. `feat(madqc):` **the driver** (renames madqc_v2) — one input deck drives
   SCF → response; `engine v2` fails with a migration hint; legacy
   `quadratic true` maps to hyperpolarizability; `calc_info.json` is written
   rank-0/atomically with failure records; the `.out` summary consumes the new
   property schema.
4. `feat(world,mra):` WorldProfile JSON emitter (machine-readable per-phase
   profile + run context) and small robustness fixes (plotdx empty-cell guard,
   PathManager absolute paths).
5. `docs:` ten-chapter parallel-runtime developer guide + performance-model
   companion.
6. `test(madqc):` regenerated h2o α+β regression golden (modern engine schema).

### Validation
- h2o α/β vs Dalton-derived references (α 0.02%, β 0.33% Kleinman-consistent);
  α bit-stable to 12 digits across every refactor gate.
- c6h6 (21 occupied) multi-frequency α reference: all channels converged;
  ᾱ(ω=0) ≈ 68.4 au with clean dispersion.
- State-parallel FD: α matches single-world to 3.5e-7; measured 2–2.6× at
  2–4 subworlds with β bit-exact to 1.1e-10.
- Tiered CTest: `unit` (pure logic, runs anywhere), `response-small`
  (self-contained synthetic-GS solver tests), `response-fixture` (archive
  regressions; registered only when `MADNESS_RESPONSE_FIXTURE_DIR` is set, so
  CI without fixtures stays green).

### Notes for reviewers
- The series is sliced for review clarity; intermediate commits are not
  individually configurable — the tip is the validated tree (9/9 gate verdicts
  on this exact base, including upstream's de-vendored nlohmann_json).
- Excited states ship as experimental (documented in
  `src/apps/molresponse/docs/FEATURE_STATUS.md`); full-tensor per-atom Raman is
  roadmapped (`src/apps/madqc/ARCHITECTURE_ROADMAP.md`) with the archived v2
  tensor as its validation target.


ahurta92:
CI note: the two **Release Threads** failures were a missing header, not a code problem — `build_response_ground_state.hpp` is matched by the `.gitignore` pattern `build*` (intended for build directories), so it was silently excluded from the branch while present in the validated working tree. Fixed in 2c8d562ce, which also removes eight orphaned `molresponse_v2` headers inherited from master (their consumers were already deleted in the first commit of this series).

Heads-up for upstream, independent of this PR: any future source file named `build_*.hpp` will hit the same silent exclusion unless the pattern is tightened (e.g. `build*/`).

ahurta92:
CI follow-up: after the header fix, ubuntu went fully green (build + tests). The remaining macos *Release Threads* failure was one test — `test_state_metrics` asserts RSS > 0, but the meter read `/proc/self/statm`, which doesn't exist on macOS. Fixed in 33eac1e2b with a `mach task_info` branch, so the memory meter now reports real RSS on both platforms. 96/97 → expecting 97/97.

fbischoff (approved):

PR #705, opened 2026-07-09
build: discover-or-fetch GoogleTest instead of bundling an ancient copy

### LINKED ISSUE #703: new json old gtest conflicts
Pulling the new json PR makes gtest compilation fail, Claude comments below. Should we also bump up gtest similar to json?

The errors are all in the bundled googletest (src/madness/external/gtest/gtest-all.cc), which is being compiled against the wrong gtest headers. The compile line for MADgtest on node26 is:

-I…/external/gtest  -I…/src  -I/users/flbi/miniconda3/include  -I…/external/gtest/include

/users/flbi/miniconda3/include is searched before the bundled gtest's own include/ dir, and miniconda has a modern googletest (≥1.10) installed there. The bundled gtest is ancient — its sources use testing::internal::string, Int32, and TypedTestCasePState, all of which were removed/renamed in modern gtest headers. Hence exactly the errors you see ('Int32' has not been declared, no declaration matches …TypedTestSuitePState…).

How miniconda's include dir gets there: the top-level CMakeLists.txt:709 does a global

include_directories($<TARGET_PROPERTY:nlohmann_json::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)

and on node26 find_package(nlohmann_json) resolved to the conda package (nlohmann_json_DIR=/users/flbi/miniconda3/share/cmake/nlohmann_json), whose interface include dir is /users/flbi/miniconda3/include. That global include is applied to every target — including the bundled gtest — and happens to drag conda's gtest headers along. (Your Mac build is unaffected because there nlohmann_json isn't found in a prefix that also contains gtest.)

evaleev:
Fixes #703.

## Problem

The bundled GoogleTest under `src/madness/external/gtest` is pre-1.10, and its header directory is added to the compile line at the **lowest** include priority — it comes in as a target-level `target_include_directories(MADgtest PUBLIC …)`, which CMake appends *after* every directory-scoped `include_directories()`. So any earlier `-I` carrying a modern `gtest/gtest.h` shadows it, and the ancient `gtest-all.cc` fails to compile against ≥1.10 headers (removed APIs: `Int32`/`UInt32`, `testing::internal::string`, `TypedTestCasePState`).

This is a **latent** fragility, not a json regression. PR #701 was the trigger (it put a conda prefix on the global include path), but the same clash fires for any modern gtest reachable through a globally-added dependency prefix — a conda compiler's `CPPFLAGS=-I$PREFIX/include`, MPI/BLAS found in that prefix, `CMAKE_PREFIX_PATH`, etc.

## Fix

Replace the bundled copy with `cmake/modules/FindOrFetchGTest.cmake`, following the repo's existing `FindOrFetch*` convention:

1. reuse a GTest target a parent project already defined,
2. `find_package(GTest)` (CONFIG then MODULE),
3. `FetchContent` googletest (`v1.14.0` by default, overridable via `MADNESS_FETCH_GTEST_TAG`).

The result is exposed under the same `MADgtest` target — **gtest only, no `gtest_main`**, matching prior behavior (test drivers keep their own `main()`), so the six unit-test link sites are unchanged. `MADNESS_HAS_GOOGLE_TEST` stays unconditionally `1`. The bundled tree is deleted so the shadowing hazard is gone entirely.

Test sources using macros modern gtest deprecates were modernized (`TYPED_TEST_CASE` → `TYPED_TEST_SUITE`, `INSTANTIATE_TEST_CASE_P` → `INSTANTIATE_TEST_SUITE_P`), which `-Werror` would otherwise reject.

## Verification

- **Fetch path, standalone:** fetched v1.14.0, built `libgtest.a`, compiled + ran a TU through `MADgtest` (own `main`) — passed.
- **In-tree** (no system/conda gtest present → fetch path exercised, `-Werror` on): configure clean; `test_stack` (24 tests) and `test_tensor` (30 tests, typed suite) built and passed.

## Notes / follow-ups

- `test_gentensor.cc` compiles only under `-DENABLE_GENTENSOR=ON`, and `test_mra6.cc` is not in any source list (orphan). Their edits are mechanically identical to the verified `test_tensor.cc` rename but were not compiled in a GENTENSOR build here.
- Separately worth cleaning up (not required for this fix): `nlohmann_json` should be consumed via a `nlohmann_json::nlohmann_json` link target rather than a global `include_directories()`, so a dependency prefix never lands on the global include path again.


evaleev:
@fbischoff please try

fbischoff:
fix works for me

PR #717, opened 2026-07-27
Clean up C++ version checks

### LINKED ISSUE #713: C++17 and more

devreal:
Allow newer versions than C++ and require at least C++-17. Remove workaround for CheckCXXSourceCompiles since starting from 3.8.0 it respects CMAKE_CXX_STANDARD (see https://gitlab.kitware.com/cmake/cmake/-/work_items/15361) We already require at least CMake 3.12.0.

These checks fail if a consuming project requires C++ newer than C++-20.

Fixes #713


evaleev:
while at this: why not C++20? I get whiplash from going to madness source and not being able to use concepts/ranges.

@fbischoff @robertjharrison

evaleev (approved):

PR #729, opened 2026-08-06
Pr scf perf

hborchert:
## What this does

Three performance changes in the SCF plus a load-balance fix. None alter results. The
kinetic-energy matrix carries most of the code.

**1. Pre-stage neighbor coefficients instead of fetching them.**

Differentiating a function fetches the coefficients of each local node's neighbors along the
axis. These fetches are cross-rank and latency-bound. The kinetic-energy matrix differentiates
`n` orbitals along three axes and issues them in bulk. Profiling put its differentiation at
**65–87% idle**.

The neighbor relation is symmetric and ownership is a pure function of the key. A producer can
therefore compute its own consumers without asking. Node `M` is the neighbor of `M-1` and `M+1`.
Its owner pushes it to the owners of those two keys, one batched message per destination.
`find_neighbor` then serves the node locally.

The cache reproduces what `sock_it_to_me` returns for the same key:

| cache state | meaning | `sock_it_to_me` |
|---|---|---|
| present, has data | same-level leaf | coefficients |
| present, empty | interior node, descend | empty |
| absent | neighbor is coarser | walks up to the coarse ancestor |

The third case is not covered. A coarse ancestor pushes only to the neighbors at its own level.
A miss therefore falls through to the existing fetch. Misses are always correct. `do_diff1`,
`do_diff2` and `matrix_inner` are unmodified.

**2. Spawn the per-node differentiation tasks from the task pool.**

`FunctionImpl::diff` submits one task per local node from a serial loop on the main thread. That
loop becomes the limiting term at high thread counts. It also issues the neighbor fetches
serially, so their latencies do not overlap. `taskq.for_each` over the same nodes produces the
same tasks. Only the spawning is parallel. This is the cost that remains once the fetches are
gone, so it composes with (1).

**3. Tile `V*psi` only at the tight protocols.**

Tiling the local-potential multiplication bounds the transient memory of the products. That peak
only threatens at high precision. The tiling costs `nmo/ntile` fences and a truncation of every
product at every protocol. The untiled path was the default for years. Its fence count does not
grow with `nmo`.

**4. Cost the load-balance root node like an ordinary node.** See below.

## Load-balance root node

`lbcost` charged the level-0 node 100× a normal node. `add_tree` accumulates that once per
function. `key0`'s cost therefore grew as `~100*(leaf+parent)*n_funcs`. On a balance over many
light trees `key0` outweighs every real subtree. The bin-packer gives it to one rank, reads that
rank as the most loaded, and starves it of work. One rank then sits idle. The loss is a flat
`1/nproc`.

This is called out separately because `lbcost` is generic. The kinetic-energy matrix balances
over `n` orbital trees and hits it squarely. Every other `load_balance` caller was paying it to
some degree. Only tree ownership changes. Results are unaffected.

## Commits

| commit | what |
|---|---|
| `5f628fade` | cost the load-balance root node like an ordinary node |
| `8fdf52350` | accumulate the tiled `V*psi` products with `gaxpy` |
| `5883389a5` | tile `V*psi` only at the tight protocols |
| `1b62da8ff` | pre-stage neighbor coefficients for the derivative operator |
| `6d5d37d81` | unit test for the neighbor halo |
| `1a1045316` | submit the derivative's per-node tasks from the task pool |

6 commits, +330/−9 over 7 files. 194 insertions are the new test.

## Interface changes

**1. New optional API on `FunctionImpl`.** `halo_enable`, `halo_clear`, `halo_enabled`,
`halo_size`, `receive_halo`, `halo_probe`. The cache is a `unique_ptr` and is null until a caller
stages one. A function that never stages a halo carries nothing. Consulting it costs one pointer
test. Nothing outside the kinetic-energy matrix stages one.

**2. Staging is collective and fence-ordered.** Every rank must call `halo_enable` and fence
before any rank stages. A push arrives as a task and cannot allocate the cache without racing
that call. A second fence separates staging from differentiating. `receive_halo` asserts that the
cache exists, so a violation fails loudly. `SCF::kinetic_energy_matrix` owns these fences.

**3. `DerivativeBase` gains `parallel_submit_`.** Per instance and off by default. Ordinary
derivative applications keep the serial submission. The kinetic-energy matrix opts in. Its
gradient operators are used nowhere else.

**4. Transient memory.** The staged coefficients are the remote-neighbor surface. All three axes
are staged before the single fence. The union of the three surfaces is live until the derivatives
are taken, then freed. The cost scales with the surface of the trees, not the volume.

No checkpoint-format change. No default behaviour change outside the kinetic-energy matrix.

## Tests

New `src/madness/mra/test_halo.cc` (`short`). Registered at 1 rank, and at 2 via
`add_mpi_tests`. The halo only carries data between ranks, so a single-rank run cannot exercise
it.

Both features must be numerically inert. That is a property of the code rather than of a
calculation, so the test asserts it directly. It compares the derivative taken with a staged halo
against the same derivative taken without one. The two perform identical arithmetic, so agreement
is required to round-off rather than to a tolerance. Three cells: halo alone, pool submission
alone, and both.

The operands are cusped Slater functions at four centres and widths. The tree is then deep and
unevenly refined. Neighbors resolve as same-level leaves, as interior nodes, and as coarser
ancestors the cache does not cover.

Two further assertions. The halo must be non-empty, so a staging path that moved nothing cannot
pass. Clearing the halo must restore the plain path, so a stale cache cannot leak into a later
derivative.

Mutation-tested. Pushing the consumer's key instead of the producer's is a plausible off-by-one
in the symmetry argument. It is caught at `3.8e0` against a `1e-13` bound. The submission-only
cell still reports `0.0`, so the failure localizes.

## PR checklist

| item | answer |
|---|---|
| raw MPI calls | none added; staging goes through `WorldObject::task` |
| blocking MPI collectives | none added; the two new fences are `World::gop.fence()` |
| implicit default World | none added; `stage_halo` uses the operator's `World&` |
| non-collective container construction | none. Staging does add a collective ordering requirement, see interface change 2 |
| `Function` mutation from user threads | no. The halo is written from the main thread and from tasks, and read only after a fence |
| tree-state preconditions | unchanged. The caller reconstructs as before, and the halo is keyed by node |
| GENTENSOR coverage | `coeffT` may be a `GenTensor` and now travels in a new message payload. `MADchem` and `testsuite` compile clean under `-DENABLE_GENTENSOR=ON` (Debug, C++20), and `test_halo` runs the path in every build |
| load-balance ownership | changed for every `load_balance` caller, see above |

## Verification

- the new test at 1, 2, 4 and 8 ranks. Agreement is exactly `0.0` in every cell
- `testsuite` green, serial and under MPI
- the `short|medium` suite, no new failures
- `-DMADNESS_WERROR=ON` build clean
- `-DENABLE_GENTENSOR=ON` Debug C++20 build clean
- `clang-tidy` reports no warnings in the added header code. The new test matches the count of
  the merged `test_mul_sparse.cc`. Two `narrowing-conversions` hits on a changed `SCF.cc` line
  match master's own idiom two lines above
- `moldft` on the three-protocol ladder, HF and LDA. LDA matches the pre-change reference at all
  three protocols. HF matches the two tighter ones and moves by `1e-8` at the loosest, from the
  `V*psi` truncation placement. Every protocol still tightens on the one before

### DIFF
--- a/src/madness/chem/SCF.cc
+++ b/src/madness/chem/SCF.cc
@@ -1414,7 +1414,10 @@ vecfuncT SCF::apply_potential(World& world, const tensorT& occ,
 
     // compute Vpsi and truncation
     START_TIMER(world);
-    const bool tile_Vpsi = true;
+    // Tiling bounds the products' transient memory, which only threatens at high precision, and
+    // costs nmo/ntile fences plus a truncation of every product. The untiled path was the default
+    // for years; take it below the tight protocols.
+    const bool tile_Vpsi = FunctionDefaults<3>::get_thresh() < 1.e-7;
     size_t min_tile = 10;
     size_t ntile = std::min(amo.size(), min_tile);
     if (!molecule.parameters.pure_ae()) {
--- a/src/madness/chem/SCF.cc
+++ b/src/madness/chem/SCF.cc
@@ -692,6 +692,8 @@ distmatT SCF::kinetic_energy_matrix(World& world, const vecfuncT& v) const {
     START_TIMER(world);
     reconstruct(world, v);
     END_TIMER(world, "KEmat reconstruct");
+    // these operators serve only this matrix, which is the many-tree case the pool submission helps
+    for (int axis = 0; axis < 3; ++axis) gradop[axis]->parallel_submit_ = true;
     START_TIMER(world);
     // Pre-stage the neighbor coefficients the ranks owe each other, so differentiating serves them
     // locally instead of fetching one at a time. The halo is keyed by node, so one covers all axes.
--- a/src/madness/chem/SCF.cc
+++ b/src/madness/chem/SCF.cc
@@ -1425,14 +1425,11 @@ vecfuncT SCF::apply_potential(World& world, const tensorT& occ,
                 size_t iend = std::min(ilo+ntile,amo.size());
                 vecfuncT tmpamo(amo.begin()+ilo,amo.begin()+iend);
                 auto tmpVpsi = mul_sparse(world, vloc, tmpamo, vtol);
-
-                //truncate tmpVpsi
                 truncate(world, tmpVpsi);
 
-                //put the results into their final home
-                for (size_t i = ilo; i<iend; ++i){
-                    Vpsi[i] += tmpVpsi[i-ilo];
-                }
+                // one gaxpy per tile instead of two fences per orbital, as the untiled branch does
+                vecfuncT Vpsi_tile(Vpsi.begin()+ilo, Vpsi.begin()+iend);
+                gaxpy(world, 1.0, Vpsi_tile, 1.0, tmpVpsi);
             }
             END_TIMER(world, "V*psi");
         } else {
--- a/src/madness/chem/SCF.cc
+++ b/src/madness/chem/SCF.cc
@@ -693,11 +693,20 @@ distmatT SCF::kinetic_energy_matrix(World& world, const vecfuncT& v) const {
     reconstruct(world, v);
     END_TIMER(world, "KEmat reconstruct");
     START_TIMER(world);
+    // Pre-stage the neighbor coefficients the ranks owe each other, so differentiating serves them
+    // locally instead of fetching one at a time. The halo is keyed by node, so one covers all axes.
+    for (int i = 0; i < n; ++i) v[i].get_impl()->halo_enable();
+    world.gop.fence();
+    for (int axis = 0; axis < 3; ++axis)
+        for (int i = 0; i < n; ++i) gradop[axis]->stage_halo(v[i].get_impl().get());
+    world.gop.fence();
     vecfuncT dvx = apply(world, *(gradop[0]), v, false);
     vecfuncT dvy = apply(world, *(gradop[1]), v, false);
     vecfuncT dvz = apply(world, *(gradop[2]), v, false);
     world.gop.fence();
     END_TIMER(world, "KEmat differentiate");
+    // all three axes' surfaces are live here, since all three are applied before the fence above
+    for (int i = 0; i < n; ++i) v[i].get_impl()->halo_clear();
     START_TIMER(world);
     compress(world, dvx, false);
     compress(world, dvy, false);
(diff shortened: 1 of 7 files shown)


fbischoff (approved):

src/madness/mra/derivative.h
```diff
@@ -237,13 +242,40 @@ namespace madness {
             }
         }
 
+        /// Push f's remote same-level neighbor coefficients to the ranks that will need them.
+
+        /// Node M is the neighbor of M-1 and M+1, so its owner pushes it to the owners of those keys
+        /// without being asked, one batched message per destination. Requires `halo_enable()` and a
+        /// fence first, then a fence before differentiating; `halo_clear()` frees the result.
+        void stage_halo(const implT* f) const {
```
devreal, line 250: This is specific to differentiation. Are there other operations that could use a halo optimization but with a different pattern? Should `stage_halo` be `stage_diff_halo`?

devreal: And should it have a `bool fence` argument to make it clear in the signature that it requires a fence?

src/madness/chem/SCF.cc
```diff
@@ -692,12 +692,23 @@ distmatT SCF::kinetic_energy_matrix(World& world, const vecfuncT& v) const {
...
+    // these operators serve only this matrix, which is the many-tree case the pool submission helps
+    for (int axis = 0; axis < 3; ++axis) gradop[axis]->parallel_submit_ = true;
     START_TIMER(world);
+    // Pre-stage the neighbor coefficients the ranks owe each other, so differentiating serves them
+    // locally instead of fetching one at a time. The halo is keyed by node, so one covers all axes.
+    for (int i = 0; i < n; ++i) v[i].get_impl()->halo_enable();
+    world.gop.fence();
+    for (int axis = 0; axis < 3; ++axis)
+        for (int i = 0; i < n; ++i) gradop[axis]->stage_halo(v[i].get_impl().get());
+    world.gop.fence();
```
devreal, line 704: It seems we could use a vmra function for this.

PR #731, opened 2026-08-07
Jm/memory profile

JonathonMisiewicz:
Adds some technology to MADNESS to assist in memory profiling.

Ed, I know there were some comments you wanted me to address. Even before I rebased, I couldn't find them on any of these commits. I'm happy to do them if you can point me to them.


evaleev (commented):
## Review

### Blockers

**1. `src/madness/mra/stacktrace_util.h` is missing from the PR.** `funcimpl.h` and `memory_measurement.h` both `#include <madness/mra/stacktrace_util.h>`, and `StacktraceCapture::capture()` is called in two `FunctionImpl` ctors — but the file exists neither on this branch nor on master (`git ls-tree <head> src/madness/mra/ | grep -i stack` → nothing). The feature cannot compile at all with `-DMADNESS_ENABLE_MEM_PROFILE_STACKTRACE=ON`. CI doesn't catch it because the option defaults to OFF. Probably lost in the rebase.

**2. `src/madness/mra/vmra.h` — leftover debug instrumentation in a hot path.** `truncate(World&, vector<Function>&)` has ~179 call sites across `src/madness` and `src/apps`. The added code runs unconditionally, not under any profiling macro:

- Two extra `world.gop.fence()` calls per invocation — global synchronization added to one of the most-called vectorized routines, in every build configuration.
- `malloc_trim(0)` twice per invocation — returns arena memory to the OS; expensive and defeats allocator reuse.
- `print_meminfo(..., "input.meminfo")` — the hardcoded `"input.meminfo"` prefix is clearly from one specific debugging session. (The body is a no-op unless `WORLD_MEM_PROFILE_ENABLE`, but the surrounding fences/trim are not.)
- `World::find_instance(SafeMPI::COMM_WORLD)->rank()` dereferences without a null check, and uses the COMM_WORLD rank rather than `world.rank()`.

This should come out entirely, or be gated behind `MADNESS_ENABLE_MEM_PROFILE_STACKTRACE` / `WORLD_MEM_PROFILE_ENABLE`. The block is also indented 7 spaces against the surrounding 8.

**3. Boost is linked to the wrong target, and the export breaks.** In `src/madness/mra/CMakeLists.txt`, `target_link_libraries(MADmra PUBLIC Boost::headers)` — but `MADmra` has no sources; it's assembled from `$<TARGET_OBJECTS:MADmra-obj>`. `AddMADLibrary.cmake` forwards only `$<TARGET_PROPERTY:MADmra,INCLUDE_DIRECTORIES>` to the object library, which does *not* include usage requirements pulled in transitively from `Boost::headers`. So `MADmra-obj` — the target that actually compiles `funcimpl.h` — never gets Boost's include path. Link against `MADmra-obj` as well.

Separately, `PUBLIC` on an exported target puts `Boost::headers` and `${CMAKE_DL_LIBS}` into `install(EXPORT madness …)`, but `cmake/madness-config.cmake.in` has no `find_dependency(Boost)` — install-tree consumers get an undefined imported target.

### Significant

**4. Global visibility settings leak into parent projects.** `CMakeLists.txt` sets `CMAKE_CXX_VISIBILITY_PRESET` and `CMAKE_VISIBILITY_INLINES_HIDDEN` as **cache** variables. Cache entries are global, so when MADNESS is consumed via `add_subdirectory`/`FetchContent` (TiledArray, MPQC), this silently changes visibility for the *parent's* targets too. If the intent is to protect MADNESS's own RTTI, set these as directory-scope (non-cache) variables, or set the properties on the `MAD*` targets. Note MADNESS doesn't set `-fvisibility=hidden` anywhere today, so this is defensive against a parent that does — which is exactly the case where clobbering the parent's setting is most harmful.

**5. `MADNESS_ASSERT` should be `MADNESS_CHECK`.** `world.cc` asserts `World::default_world->id() == 0`, and `memory_measurement.h` now depends on that fact (`all_worlds.push_back(0)`). `MADNESS_ASSERT` is debug-only and compiled out in release depending on `ASSERTION_TYPE`, so in release the invariant the new code relies on goes unchecked.

The invariant itself does hold: `initialize()` calls `initialize_world_id_range(comm.Get_rank())` before constructing the default world, rank 0's range starts at 0, and `_id` is `Bcast` from rank 0 — including across repeated initialize/finalize cycles. Consider exposing it as `World::default_world_id()` rather than a bare literal `0` at the call site.

**6. Stacktrace symbolization timing.** `creation_stacktrace_` is a `std::string`, so if `StacktraceCapture::capture()` symbolizes eagerly, every `FunctionImpl` construction forks `addr2line` once per frame. With `MADNESS_MEM_PROFILE_STACKTRACE_DEPTH=0` (unlimited, the default) that would be unusable even for profiling. Storing the `boost::stacktrace::stacktrace` object and symbolizing lazily in `print_stacktrace_details` would avoid it. Can't confirm which it does — see blocker 1.

**7. `BOOST_STACKTRACE_USE_ADDR2LINE` is defined in the installed `config.h`.** `config.h` is included by essentially every MADNESS header, so this forces the addr2line backend on any downstream TU that uses Boost.Stacktrace for its own purposes. Better confined to `stacktrace_util.h`.

Also worth verifying: Boost stringizes `BOOST_STACKTRACE_ADDR2LINE_LOCATION` internally, so passing an already-quoted `"@MADNESS_STACKTRACE_ADDR2LINE_LOCATION@"` likely yields a path containing literal quote characters and the exec fails silently. I couldn't check this directly — no Boost.Stacktrace headers on the machine I reviewed from.

And on macOS the relevant tool is `atos`, not GNU `addr2line`; the CMake comment mentions "addr2line/atos" but only ever searches for `addr2line`.

**8. The report table changes shape when the feature is on.** Adding `func_id` to `MemKey` makes every row a single `FunctionImpl`, so `num_functions` is always 1 and the table grows to one line per function per rank — the aggregate view is lost in exactly the build where you'd want both. Relatedly, in `print_stacktrace_details`, `stacktrace_count[trace] += memval.num_functions` sums over per-rank rows, so "Number of functions" reports *nranks × functions*, not the function count.

### Minor

- `CMakeLists.txt` — whitespace-only reindent of `####### preferred task backend`, unrelated to this PR.
- `set(MADNESS_ENABLE_MEM_PROFILE_STACKTRACE OFF)` in the two failure branches sets a normal variable shadowing the cache option; the cache stays `ON`, so the warning re-fires on every reconfigure and `add_feature_info` (registered earlier) still reports the feature enabled.
- The Boost-detection block re-derives Boost availability instead of using `MADNESS_HAS_BOOST`, which `external/boost.cmake` already sets.
- `src/madness/mra/CMakeLists.txt` — `if(NOT WIN32)` guard; MADNESS is POSIX-only, no Windows branches needed.
- `config.h.in` — `MADNESS_MEM_PROFILE_STACKTRACE_DEPTH` is defined unconditionally in the public header even when the feature is off.
- `memory_measurement.h` `operator<`: the `#else return false;` branch is correct (equivalent to the old `lhs.DIM<other.DIM` when DIMs are equal), just worth a comment since it reads oddly.
- `test_memory_measurement.cc` asserts that distinct call sites yield distinct traces, which is inlining/LTO-dependent despite the `__attribute__((noinline))` — could be flaky at `-O3`.

### Things that look good

- Replacing `World::get_default().id()` with the default-world id is the right direction for `-DMADNESS_DISABLE_WORLD_GET_DEFAULT=ON`.
- No raw `MPI_*` calls and no blocking MPI collectives; `gop.concat0` for the stacktrace gather is the correct idiom.
- `uniqueidT` has `operator<`, `operator bool`, and archive support, so the `std::map<uniqueidT, std::string>` and `concat0` usage is sound.
- Checkpoint format (`Function::save`/`load`) and `FunctionDefaults` are untouched.

The two to fix first are the missing `stacktrace_util.h` and the `vmra.h` instrumentation — the latter affects everyone regardless of build flags.

PR #739, opened 2026-08-11
Pr exchange p2p

hborchert:
## What this does

Reworks the symmetric HF exchange so that the orbitals are stored once as batches owned by
particular ranks, each task runs where its data already is, and the batch it will need next is
requested while it computes. Adds the runtime and macrotask-framework machinery this needs, and a
consumer that uses it.

Reachable as `hfexalg multiworld`. **moldft's default is `multiworld_row`, which is untouched**, so
no existing calculation changes behaviour unless it asks for the new path.

**1. `world/cloud.h` — owner-pinned batches and a point-to-point transport.**

- a second container holds batches pinned to an owning rank; a requester asks the owner directly
  and gets the bytes back, resolving locally when it is the owner
- both endpoints are posted from comm-thread handlers, so neither queues behind saturated workers
- a transfer is split into 1 GiB messages, because an MPI byte count is an `int` and batches are
  measured above 2 GiB
- batch records are released with the rest of the cloud, and their size is reported alongside it

**2. `mra/macrotaskq.h` — optional task hooks and owner-pinned scheduling.**

- the hooks are stated as detection traits in the shape `madness::archive` uses
  (`has_store_batches_v` and friends): a task opts in by declaring a member, and one that declares
  nothing is unaffected
- tasks may name an owner (`owner_hint`), move their own operands (`handles_own_data_movement`),
  accumulate their own results (`accumulates_own_output`), and release what they cached (`cleanup`)
- when every task names an owner and there is one subworld per rank, each rank walks the queue and
  runs what it owns instead of asking a scheduler for work
- a task that accumulates its own output does so without fencing, so the subworld is quiesced once
  before its buffer is drained

**3. `chem/exchangeoperator.*` — the symmetric path.**

- the orbitals are split once; the task grid is the lower triangle over that split, and every task
  is placed on a rank that already holds one of its two batches, so placement costs no transfer
- tiles are computed one row at a time, so a tile holds one row of intermediates rather than all of
  them
- a task requests its successor's batch before computing its own
- results are summed per subworld, and per node before they reach the universe, so only one rank per
  node scatters between nodes
- tasks are placed by their measured cost from the previous application, because screening leaves
  the tiles uneven and counting them cannot see that

## Commits

| commit | what |
|---|---|
| `f583abaa9` | give `io_redirect` a discard mode and a target directory |
| `380b7895e` | owner-pinned batch storage and p2p fetch in the cloud |
| `ffb4a0ffd` | test the cloud's batch store and p2p fetch |
| `439309835` | optional task hooks and owner-pinned scheduling in the macrotask queue |
| `63523eafd` | let a task prefetch its next batch and accumulate its own output |
| `f7d2a678f` | write per-task output only at high print level |
| `334ebb81a` | exercise the macrotask queue's owned-task scheduling path |
| `7a602ef93` | a process CPU-time reader in `timers.h` |
| `5e572f290` | owner assignment for the symmetric exchange task matrix |
| `56444d6ff` | make the round-robin assignment linear in the task count |
| `c1c631e04` | batch record keys and a bounded batch cache |
| `a4ac84028` | stream the symmetric exchange tile kernels |
| `d755d0920` | pin tasks to the ranks owning their batches |
| `75228867b` | fetch operands from the rank that owns them |
| `cde72883e` | report batch fetches at the scope they are read at |
| `dfe9b77a0` | sum tile results per subworld, and per node before the universe |
| `a288fb3d2` | expose the batching and accumulation settings |
| `584d7a6e4` | request the next batch while the current task computes |
| `9ad8cc514` | place tasks by their measured cost |
| `6c7ba86b7` | release the finalize reducers while their world still exists |
| `fae1c1208` | release and report the cloud's owner-pinned batches |
| `54403ab0e` | per-task profiles behind `MAD_EXCH_TASK_PROFILE` |
| `93386add9` | make the finalize's destination check always on |
| `b35bde62b` | state the optional task hooks as detection traits |
| `4bf8a9993` | split `run_all` into the phases it runs |
| `b1d2b75ed` | split `run` into the phases it runs |
| `326adfd6e` | give the profiler the stage timings and the cost matrix |
| `89af8dcd1` | split batch transfers into 1 GiB MPI messages |

28 commits, +2687/−140 over 15 files.

## Interface changes

Three parameters are added, all for `hfexalg multiworld`:

- `hfex_batch_granularity` (default 1) — batches of orbitals per rank
- `hfex_local_accumulation` (default 2) — 1 gathers per subworld, 2 also gathers per node first and
  degrades to 1 on a single node
- `hfex_cost_aware_assign` (default true) — place tasks by measured cost rather than by counting them

The existing `memory` keyword is untouched, and no checkpoint format changes.

## Tests

`madness/test/chem/test_exchange_owner_assign` and `test_exchange_batch_cache` are new and cover the
placement and the batch cache as pure functions: that every task lands on a rank holding one of its
batches, that the placement is deterministic so ranks agree without communicating, that cost-aware
placement lowers the peak load on an uneven cost model, and that batches a rank owns are never
evicted while fetched ones are bounded. `mra/test_cloud` gains cases for the batch store, the
point-to-point fetch, a fetch of a record no rank holds, and a transfer split across several MPI
messages. `mra/test_vectormacrotask` gains a task that names owners, so the owned-walk scheduling
path is exercised.

## PR checklist

| item | answer |
|---|---|
| raw MPI calls | none. The batch transport's `Isend`/`Irecv` go through `World::mpi` (`SafeMPI`) |
| blocking MPI collectives | none added |
| implicit default World | none added. `store_batches` does point `FunctionDefaults`' pmap at the subworld for the duration of a cross-world copy and restores it, because a copy picks its pmap from the target world |
| non-collective container construction | none. The batch container, its pmap and `BatchTransport` are constructed with the cloud, in the universe |
| threaded-BLAS assumptions | none. Parallelism is macrotasks over the task queue |
| `Function` mutation from user threads | no. Kernels run in tasks; the comm-thread handlers touch only serialized bytes and MPI requests, never a `Function` |
| tree-state preconditions | unchanged; the streamed kernels take the same states as master's |
| GENTENSOR coverage | builds and tests clean under `-DENABLE_GENTENSOR=ON` |

### DIFF
--- a/src/madness/chem/CMakeLists.txt
+++ b/src/madness/chem/CMakeLists.txt
@@ -164,7 +164,8 @@ if(BUILD_TESTING)
   SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
   # The list of unit test source files
   set(CHEM_TEST_SOURCES_SHORT test_pointgroupsymmetry.cc test_masks_and_boxes.cc
-  test_qc.cc test_MolecularOrbitals.cc test_BSHApply.cc test_projector.cc)
+  test_qc.cc test_MolecularOrbitals.cc test_BSHApply.cc test_projector.cc
+  test_exchange_owner_assign.cc)
   set(CHEM_TEST_SOURCES_LONG test_localizer.cc test_ccpairfunction.cc test_low_rank_function.cc)
   if (TARGET Libxc::xc)
     list(APPEND CHEM_TEST_SOURCES_SHORT test_dft.cc )
--- a/src/madness/chem/CMakeLists.txt
+++ b/src/madness/chem/CMakeLists.txt
@@ -165,7 +165,7 @@ if(BUILD_TESTING)
   # The list of unit test source files
   set(CHEM_TEST_SOURCES_SHORT test_pointgroupsymmetry.cc test_masks_and_boxes.cc
   test_qc.cc test_MolecularOrbitals.cc test_BSHApply.cc test_projector.cc
-  test_exchange_owner_assign.cc)
+  test_exchange_owner_assign.cc test_exchange_batch_cache.cc)
   set(CHEM_TEST_SOURCES_LONG test_localizer.cc test_ccpairfunction.cc test_low_rank_function.cc)
   if (TARGET Libxc::xc)
     list(APPEND CHEM_TEST_SOURCES_SHORT test_dft.cc )
--- a/src/madness/chem/CalculationParameters.h
+++ b/src/madness/chem/CalculationParameters.h
@@ -69,6 +69,7 @@ struct CalculationParameters : public QCCalculationParametersBase {
 		initialize<std::string> ("xc","hf","XC input line");
 		initialize<std::string> ("hfexalg","multiworld_row","hf exchange algorithm; multiworld stores the orbitals once as batches and has each task fetch the two it needs from the rank holding them, which bounds memory but needs one subworld per rank",{"multiworld","multiworld_row","fetch_compute","smallmem","largemem"});
 		initialize<long>  ("hfex_batch_granularity",1,"hfexalg=multiworld: batches of orbitals per rank (>=1). 1 is the coarsest and has the lowest peak memory; raising it makes tasks smaller and easier to balance at the cost of more transfers");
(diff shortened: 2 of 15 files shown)


fbischoff (approved):
some follow-up needed, good for now

PR #753, opened 2026-08-14
Fix: Correctness Follow-Up to PR #724

EricaCMitchell:
## Description

Four independent correctness bugs in `src/madness/mra`, all silent — wrong numbers, no exception and no warning. Three are wrong MRA norms (`norm2`/`trace` summed over a redundant tree, `norm_tree` identically zero on a compressed tree, `dnorm_tree` counting the scaling block everywhere); the fourth is an unchecked precondition in the screened-multiply kernel that lets an unassigned tensor reach `do_mul`. One commit per bug.

A follow-up review then added four more commits, covering the fallout of making two `const` read methods reconstruct, one remaining instance of the same overcount, and a test that was passing without measuring anything. See the [follow-up report](https://github.com/m-a-d-n-e-s-s/madness/pull/753#issuecomment-5299157580) below for the details and for the retraction of an earlier claim about the 6D screen.

## Details

- [X] **`norm2()`/`trace()` overcounted on a redundant tree** — [a864545](https://github.com/m-a-d-n-e-s-s/madness/commit/a8645452a54e72e2a425d98e3022f2bb2638fb41). `norm2()` bypassed the tree-state guard that `norm2sq_local()` carries and `trace()` had none, so both summed every level of a redundant tree. #724 made that reachable from `SCF.cc`, `MolecularOrbitals.cc`, and ~14 `mul_sparse` sites in `src/apps/molresponse`. The global entry points now reconstruct on entry, matching `madness::norm2(world, v)`; the local no-comm variants throw instead. Test T6.
- [X] **`norm_tree` was identically zero on a compressed tree** — [bde9acb](https://github.com/m-a-d-n-e-s-s/madness/commit/bde9acba39162dc741cbd99ec45f044c06294d97). `compress_op`'s leaf case read the norm *after* `clear_coeff()` and substituted `0.0`, which propagated to the root. Reordered to read it while the value is still in hand; `norm_tree` now means `||f||` on the box in every tree state. Test T7.
- [X] **`dnorm` counted the scaling block** — [304bd09](https://github.com/m-a-d-n-e-s-s/madness/commit/304bd0995f39f23a3471e7185f4bb65632985b48). `s0` was zeroed before measuring only under `key.level() > 0 && !nonstandard1`, so root `dnorm_tree` came out equal to `||f||` exactly. Now zeroed unconditionally to measure and restored where the stored tensor keeps it, reusing the copy already made for contiguity. Test T7 pins `sqrt(snorm^2 + dnorm_tree^2) == ||f||` at the root to 1e-12 relative, so a too-small `dnorm` fails too. Checked with `-DENABLE_GENTENSOR=ON`. **This is a consistency fix, not a change in numbers**: tracing all three non-printing readers shows none of them ever saw the inflated value — the two `get_dnorm_tree()` readers in `mulXXveca` and the `ShallowNode`/`CoeffTracker` reader both run on redundant trees, where `make_redundant_op` already zeroed `s0` unconditionally, and the `get_dnorm()` reader in `get_contraction_node_lists` is reached only through `innerXX`, whose `prepare` step runs `recompute_snorm_and_dnorm()` and overwrites whatever `compress_op` left behind. (An earlier revision of this description claimed the 6D partial inner product screen tightens; it does not. See the follow-up report.)
- [X] **`mulXXveca` could reach `do_mul` with an unassigned tensor** — [3874c62](https://github.com/m-a-d-n-e-s-s/madness/commit/3874c62240007c482a03b2dd3f3b9a6ae1c2b622). `full_tensor_copy()` was called without a `has_coeff()` check, and the "uncomputed dnorm_tree" warning did not cover it — one-shot, and it did not disable screening as its message claimed. Replaced by `MADNESS_CHECK` on the redundant-form invariants. The second right-operand fetch keeps its node-existence check conditional on `vrc[i]`: parent coefficients legitimately stand in there, and an unconditional check fires 108 times in the existing tests.
- [X] **`cross()` did not fence after `reconstruct` and `compress`** — [90d9968](https://github.com/m-a-d-n-e-s-s/madness/commit/90d9968a6fca74ee57ec2386f5b4c56ec8050d6e). `cross()` fenced only through the second call of each pair. That call does a fast return when its functions are already in the correct state. Thus it does not fence, and the tasks of the first call stay incomplete. This caused failures in `testvmra`. Each pair now uses `fence=false` and one explicit `world.gop.fence()`.

### Follow-up commits

- [X] **T6 was passing without testing anything** — [fe5ce81](https://github.com/m-a-d-n-e-s-s/madness/commit/fe5ce81cc). `norm2()` and `trace()` reconstruct as a side effect, so calling them back to back on the same operand left only the first one measuring a redundant tree. Each entry point now gets its own redundant tree, with state assertions on entry. Verified by mutation: dropping the `reconstruct` from `trace()` now fails T6.
- [X] **Reconstructing inside a `const` read is an unfenced mutation** — [d830652](https://github.com/m-a-d-n-e-s-s/madness/commit/d830652c4), [3aef9e5](https://github.com/m-a-d-n-e-s-s/madness/commit/3aef9e51d). Both entry points were previously safe to call on an operand of a `mul_sparse(..., fence=false)` that had not been fenced yet — `SCF::twoint` has exactly this shape — and the removal tasks can now strip coefficients out from under a `mulXXveca` still reading them. Fence before changing state, in `Function::norm2`/`trace` and in `norm2s`/`norm2s_T`/`norm2(world,v)`. Also names the on-demand case instead of dying in `change_tree_state`.
- [X] **`print_size` printed the same overcounted norm** — [dbf1b16](https://github.com/m-a-d-n-e-s-s/madness/commit/dbf1b161c). It is a diagnostic and must not mutate the tree, so it reports the norm only where it means something and `n/a` otherwise.
- [X] **`inner()` hid the tree state of an on-demand function** — [fed1c1a](https://github.com/m-a-d-n-e-s-s/madness/commit/fed1c1abe9e21664b32e9b276f91e2f0181db200). In the self-inner shortcut, `inner()` changed the tree state to reconstructed before it called `norm2()`. This made an on-demand function give a value in place of an error. The shortcut now calls `norm2()` directly, which does the state check.
- - [x] **`zero_norm_tree()` left `dnorm_tree` stale** — [1360307](https://github.com/m-a-d-n-e-s-s/madness/commit/1360307fe8e1f59f55e28753775b78670742dbbf). It cleared `norm_tree`, `snorm`, and `dnorm` but not `dnorm_tree`, which `mulXXveca()` reads together with `norm_tree` for screening. `broaden()` calls `zero_norm_tree()`, so a broadened tree was left with `norm_tree == 0` paired against a stale `dnorm_tree` from the last `make_redundant()`/`compress()` — latent today, but a screen against that pair would be too coarse. `dnorm_tree` is now reset to `NORM_TREE_UNCOMPUTED`, not `0.0`, since `0.0` would also pass the screening check. Test T8.

Verification (RelWithDebInfo, Apple clang, arm64, OpenMPI 5.0.9), np=1 and np=2:

```
./src/madness/mra/test_mul_sparse    # exit 0, T1–T8, no `failed` lines
./src/madness/mra/testsuite          # testsuite passed:  true, exit 0
```

Screening numbers do not move: every `err` and `C` column in the T3/T4 output is bit-identical to a baseline taken before these commits, the only diff hunks being wall-clock strings on lines that all report `passed`.

## Questions

- [X] ~~The `dnorm_tree` fix tightens screening for the 6D partial inner product (`mp2`/`cc2` paths) — worth a look from someone with those timings in hand?~~ **Resolved: it does not tighten.** No in-tree consumer reads the inflated value; see the `dnorm` bullet above and the follow-up report. No production 6D cost changes.
- [x] The `MADNESS_CHECK_THROW` added to `trace_local()`/`norm2sq_local()` turns a previously silent wrong answer into a hard failure for any caller passing a non-{compressed,reconstructed} function. I believe there are none in-tree, but it is a behaviour change for downstream consumers. The same applies to `norm2()`/`trace()`/`norm2s*()` on an **on-demand** function, which now throw instead of returning `0.0`.

## AI
- Bug-seeking, derivation, and drafting: the four bugs came out of an AI-assisted re-review of PR #724 against current master, and the T6-T8 test cases and their tolerances were drafted with AI assistance. The four follow-up commits came out of an AI-assisted review of this PR. Every finding was reproduced and every tolerance measured against real runs before being written down; the numbers quoted above are from those runs, not from the model.

### DIFF
--- a/src/madness/mra/CMakeLists.txt
+++ b/src/madness/mra/CMakeLists.txt
@@ -54,6 +54,9 @@ if(BUILD_TESTING)
   add_mpi_tests(mra test_cloud "2" "MADmra;MADgtest" "unittests;short")
   add_mpi_tests(mra test_vectormacrotask "2" "MADmra;MADgtest" "unittests;short")
   add_mpi_tests(mra test_eval "2" "MADmra;MADgtest" "unittests;short")
+  # the tree-state and norm checkpoints use gop.max/gop.sum and branch on a
+  # replicated state, so only a multi-rank run exercises them
+  add_mpi_tests(mra test_mul_sparse "2" "MADmra;MADgtest" "unittests;short")
   # the halo only carries data between ranks, so the single-rank run above cannot exercise it
   add_mpi_tests(mra test_halo "2" "MADmra;MADgtest" "unittests;short")
 
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -3057,8 +3057,12 @@ template<size_t NDIM>
 
             if (lc.size() == 0) {
                 MADNESS_CHECK(lit != left->coeffs.end());
+                // redundant form puts coefficients and computed norms on every
+                // node; without them the screen below reads garbage
+                MADNESS_CHECK(lit->second.has_coeff());
                 lnorm = lit->second.get_norm_tree();
                 ldnorm = lit->second.get_dnorm_tree();
+                MADNESS_CHECK(ldnorm < NORM_TREE_UNCOMPUTED);
                 l_is_leaf = !lit->second.has_children();
             }
             else {
@@ -3094,20 +3098,14 @@ template<size_t NDIM>
                 riterT rit = right->coeffs.find(key).get();
                 if (rc.size() == 0) {
                     MADNESS_CHECK(rit != right->coeffs.end());
+                    MADNESS_CHECK(rit->second.has_coeff());
                     rnorm = rit->second.get_norm_tree();
                     rdnorm = rit->second.get_dnorm_tree();
+                    MADNESS_CHECK(rdnorm < NORM_TREE_UNCOMPUTED);
                 }
                 else {
                     rnorm = rc.normf();
                     rdnorm = 0.0;
-                }
-
-                if (ldnorm >= NORM_TREE_UNCOMPUTED || rdnorm >= NORM_TREE_UNCOMPUTED) {
-                    static std::atomic<bool> warned{false};
-                    bool expected = false;
-                    if (warned.compare_exchange_strong(expected, true))
-                        print("WARNING: mul_sparse operand has an uncomputed dnorm_tree; "
-                              "screening is disabled for those nodes (missing make_redundant?)");
                 }
 
                 // the neglected cross terms are below threshold: multiply here (requires redundant form)
@@ -3137,7 +3135,11 @@ template<size_t NDIM>
                 std::vector< Tensor<R> > vrss(vresult.size());
                 for (unsigned int i=0; i<vresult.size(); ++i) {
                     riterT rit = vright[i]->coeffs.find(key).get();
+                    // coefficients handed down from the parent stand in for a node
+                    // that need not exist here; without them the node must exist
+                    MADNESS_CHECK(vrc[i].size() || rit != vright[i]->coeffs.end());
                     if (vrc[i].size() || !rit->second.has_children()) {
+                        MADNESS_CHECK(vrc[i].size() || rit->second.has_coeff());
                         Tensor<R> rd(cdata.v2k);
                         rd(cdata.s0) = (vrc[i].size() ? vrc[i] : rit->second.coeff().full_tensor_copy())(___);
                         vrss[i] = vright[i]->unfilter(rd);
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -4765,7 +4765,8 @@ template<size_t NDIM>
 
         void broaden_op(const keyT& key, const std::vector< Future <bool> >& v);
 
-        // For each local node sets value of norm tree, snorm and dnorm to 0.0
+        // For each local node sets norm_tree, snorm and dnorm to 0.0, and marks
+        // dnorm_tree as uncomputed.
         void zero_norm_tree();
 
         // Broaden tree
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -1442,6 +1442,25 @@ template<size_t NDIM>
 
         bool is_on_demand() const;
 
+        /// Returns true if only the leaves of this tree carry its coefficients
+
+        /// redundant and nonstandard_with_leaves keep s coefficients on the
+        /// internal nodes as well, so a sum over all nodes counts the function
+        /// more than once; the leaves alone are exactly the reconstructed tree,
+        /// which is why change_tree_state(reconstructed) is nothing but
+        /// remove_internal_coefficients() for these two states.
+        bool has_coefficients_on_leaves_only() const;
+
+        /// Returns true if summing over the local nodes yields the function
+
+        /// The precondition of norm2sq_local() and trace_local(): the tree holds
+        /// its coefficients exactly once.  reconstructed and compressed do so
+        /// outright, the two states above once the internal nodes are skipped.
+        /// redundant_after_merge is a sum that has not been collapsed yet, and
+        /// nonstandard / nonstandard_after_apply have no leaf coefficients to
+        /// single out, so neither qualifies.
+        bool has_summable_coefficients() const;
+
         bool has_leaves() const;
 
         void set_tree_state(const TreeState& state) {
@@ -5752,8 +5771,15 @@ template<size_t NDIM>
         T trace_local() const;
 
         struct do_norm2sq_local {
+            /// skip the internal nodes, cf. has_coefficients_on_leaves_only()
+            bool leaves_only = false;
+
+            do_norm2sq_local() = default;
+            explicit do_norm2sq_local(bool leaves_only) : leaves_only(leaves_only) {}
+
             double operator()(typename dcT::const_iterator& it) const {
                 const nodeT& node = it->second;
+                if (leaves_only and node.has_children()) return 0.0;
                 if (node.has_coeff()) {
                     double norm = node.coeff().normf();
                     return norm*norm;
@@ -5774,6 +5800,11 @@ template<size_t NDIM>
 
 
         /// Returns the square of the local norm ... no comms
+
+        /// Requires has_summable_coefficients(); throws otherwise, because on a
+        /// tree that carries its coefficients on more than one level the sum is
+        /// not the norm.  Internal nodes are skipped where they are the
+        /// duplicates, so no state change is needed to ask for the norm.
         double norm2sq_local() const;
 
         /// compute the inner product of this range with other
--- a/src/madness/mra/funcimpl.h
+++ b/src/madness/mra/funcimpl.h
@@ -6577,12 +6577,10 @@ template<size_t NDIM>
 
                     // make child's s coeffs if it doesn't exist or if is has no s coeffs
                     bool childnode_exists=get_coeffs().find(acc,child);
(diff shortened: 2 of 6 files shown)


hborchert (approved):

evaleev:
## Follow-up push: review findings and two retractions

Pushed on top of the reviewed state. One of the four original commits is rewritten (message only, empty code diff), and four new commits are added. Both open questions in the description are now answered — and one of them is answered *against* what the description currently claims.

### 1. `dnorm` fix does **not** tighten the 6D screen — retracting that claim

`7c56bb9` → `304bd09` is a **message-only rewrite**: `git diff 7c56bb936 304bd0995` is empty. What changed is the analysis.

The original message (and the description bullet, and open question 1) said the 6D partial inner product screen tightens because `get_contraction_node_lists` reads `get_dnorm()`. Tracing every reader shows it does not:

| reader | reached via | what actually sets `dnorm` there |
|---|---|---|
| `get_dnorm_tree()` ×2, `mulXXveca` (`funcimpl.h:3052,3091`) | `mul_sparse` | `make_redundant_op` — which already zeroed `s0` **unconditionally** (`mraimpl.h:1778`) |
| `get_dnorm()`, `get_contraction_node_lists` (`funcimpl.h:6459`) | `innerXX` only (`funcimpl.h:6358-6359`) | `prepare` runs `compute_snorm_and_dnorm` (`mra.h:2699-2700`) → `recompute_snorm_and_dnorm` zeroes `s0` itself (`funcimpl.h:355-377`), overwriting whatever `compress_op` left |
| `get_dnorm()`, `ShallowNode`/`CoeffTracker` (`funcimpl.h:780`) | `Vphi_op_NS` under `make_Vphi` | constituents are checked redundant → `make_redundant_op` again |

The remaining readers are `print`/`operator<<` only. So no in-tree consumer ever saw the inflated value: **no production 6D cost changes, and no numbers move anywhere.** This is a consistency fix — `dnorm` now means `‖d‖` in every tree state — pinned by T7 (`snorm² + dnorm_tree² == ‖f‖²` at the root).

One honest caveat, now in the commit message: the `task==2` (work-only) `innerXX` entry point relies on an earlier `task==1` call having run `prepare`. That precondition predates this PR.

**The description's `dnorm` bullet and open question 1 are stale and should be corrected to match.**

### 2. T6 was passing without testing anything (`fe5ce81`)

`norm2()` and `trace()` reconstruct as a side effect, so calling them back to back on the same operand meant only the first one measured a redundant tree — by the time `trace()` ran, `p.f` was already reconstructed and the checkpoint passed with the `trace()` reconstruct reverted.

Each entry point now gets its own redundant tree (`norm2()` on `p.f`, `trace()` on `p.g`, `p.f` put back into redundant form before `norm2s`), plus explicit state assertions on entry so a future reordering fails loudly instead of silently measuring nothing.

### 3. Reconstructing inside a `const` read is a mutation — fence first (`d830652`, `3aef9e5`)

Making `norm2()`/`trace()` reconstruct turned two read-only methods into ones that discard the interior coefficients of a redundant tree. Both are `const` and were previously safe to call on an operand of a `mul_sparse(..., fence=false)` that had not been fenced yet — `SCF::twoint` has exactly this shape. Without a fence the removal tasks land in the same queue as the multiply's, unordered, and can strip coefficients out from under a `mulXXveca` still reading them.

Both now fence before changing state, and say so in the docs the way `mul_sparse` documents leaving its operands redundant. `3aef9e5` does the same for the vector entry points (`norm2s`, `norm2s_T`, `norm2(world,v)`), factored into `detail::reconstruct_for_norm` since all three did it verbatim.

Deliberately left alone: `apply()`, `div()`, `rot()`, `refine_to_common_level()` also reconstruct their inputs, but they are declared as transformations, not as reads.

**Also an on-demand hole:** an on-demand function is neither compressed nor reconstructed, so it took the new reconstruct branch and died in `change_tree_state`'s `"unknown/unsupported current tree state"`, naming neither the function nor its state. Both entry points now check and say so. (Before this PR it returned `0.0` — worse, just quieter.)

### 4. `print_size` printed the same overcounted norm (`dbf1b16`)

`print_size()` calls `norm2sq_local()` with no tree-state guard — the last instance of the overcount `Function::norm2()` carried. It is a diagnostic and must not mutate the tree, so it cannot reconstruct its way out: it now reports the norm only where it means something and `n/a` otherwise. The tree state is replicated across ranks, so both branches stay collective.

### Verification

RelWithDebInfo, Apple clang, arm64, OpenMPI, `ENABLE_GENTENSOR=OFF`:

| | np=1 | np=2 (`MAD_NUM_THREADS=2`) |
|---|---|---|
| `test_mul_sparse` | exit 0, T1–T7 all `passed`, no `failed` lines | exit 0, no `failed` lines |
| `mra/testsuite` | `testsuite passed:  true`, exit 0 | `testsuite passed:  true`, exit 0 |

T6 re-verified by mutation: dropping the `reconstruct()` from `trace()` and rebuilding makes `test_mul_sparse` exit 1 with

```
  T6 trace  5.5683279968e+00 -> 3.3409499067e+01
  T6 trace on a redundant function                    failed
  T6 norm2/trace leave the operand reconstructed      failed
```

i.e. the checkpoint measures the fix rather than passing regardless. Reverted, rebuilt, exit 0 again.

`clang-tidy -p <build> --header-filter='src/madness/mra/(mra|vmra|mraimpl)\.h'` over `test_mul_sparse.cc`: no diagnostic on any changed line; every warning is pre-existing (`modernize-use-auto`, `bugprone-reserved-identifier`, … on untouched code).

### Behaviour changes for downstream consumers

1. `MADNESS_CHECK_THROW` in `trace_local()`/`norm2sq_local()` turns a previously silent wrong answer into a hard failure for any caller passing a non-`{compressed,reconstructed}` function. No in-tree callers; still a break for out-of-tree ones.
2. `norm2()`/`trace()`/`norm2s*()` on an **on-demand** function now throw instead of returning `0.0`.
3. `norm2()`/`trace()` on a redundant/nonstandard function now **fence** — a collective call where there previously was none on that path. Callers that reach it from divergent control flow across ranks would now hang rather than return a wrong number.
4. `print_size` prints `n/a` in place of a norm on redundant/nonstandard trees, and skips the `gop.sum`/`gop.fence` in that case. Log-scraping tools that parse that field will see `n/a`.

---

*AI assistance: the follow-up review that produced these five commits was AI-assisted, as was this report. Every claim above was re-derived against the code by hand — the reader tables in §1 are `grep`-verified call-site lists, and the test results are from the runs quoted, not from the model.*

EricaCMitchell:
Pushed two more fixes, both surfaced by `testvmra`:

  - **`cross()` never fenced** (`90d9968`) — each `reconstruct`/`compress` pair fenced only through its second call, and that call fast-returns when its functions are already in the right state. The fence was then skipped and the tasks from the first call were still in flight. Both pairs now pass `fence=false` and take one explicit `world.gop.fence()`.
  - **`inner()` masked on-demand state** (`fed1c1a`) — the self-inner shortcut forced the tree to reconstructed before calling `norm2()`, so an on-demand function returned a number instead of erroring out. The shortcut now calls `norm2()` directly and lets it do the state check.

EricaCMitchell:
This PR specifically targets issues with numerical issues caused by PR #724. A companion (PR #758) addresses an API issue introduced, but since it doesn't affect numerics it has been separated from this one.

robertjharrison:
@copilot resolve the merge conflicts in this pull request

evaleev (approved):

PR #754, opened 2026-08-15
molresponse: expose the virtual-AO excited-state guess (--es-guess / excited.guess)

ahurta92:
## What this adds

An option to pick the excited-state initial guess: `--es-guess=virtual_ao` on the test driver, or `excited.guess virtual_ao` in a madqc response deck. There's also `--es-guess-basis` to choose the AO basis it builds virtuals from (default aug-cc-pvdz). The default guess is unchanged, so existing decks behave exactly as before.

It also adds a warning when the guess "top-up" path quietly pads with extra solid-harmonic trials — before, you couldn't tell that had happened.

## Why

The default guess builds trial vectors as (solid harmonic × occupied orbital). That works well in many cases, but it can't produce a radial excitation in the totally-symmetric sector — for example He 1s→2s. And since the iteration preserves symmetry sectors, a state that's missing from the guess seems to stay missing no matter how many roots or iterations you ask for.

The `virtual_ao` guess (which already existed in the code, just wasn't reachable) builds trials from projected AO virtuals instead, so it has the radial flexibility.

## What we saw on He (same solver and settings, only the guess changed)

| guess | lowest root | finds the 2¹S? | iterations |
|---|---|---|---|
| solid harmonics (default) | 21.69 eV | no | didn't converge in 40 |
| virtual_ao | 21.13 eV | yes | 7–8 |
| random | 21.13 eV | yes | 43 |

The Gaussian-basis reference (DALTON, d-aug-cc-pVQZ) puts the 2¹S at 21.123 eV, so 21.13 looks right. The virtual_ao run also gave cleaner degeneracies. One practical payoff: with this guess we could compute the He 2¹S two-photon strength (10.29 vs DALTON's 10.60).

## Caveats

Tested mainly on a two-electron atom so far. The open-shell path still falls back to solid harmonics. Feedback on the parameter naming and defaults is very welcome.

Build + molresponse/madqc test suites pass on a fresh tree against current master.


ahurta92:
Rebased onto current master (merge-forward; CI green). This one seems to have slipped past while #755/#756 were reviewed — it's small (7 files) and is the piece that makes totally-symmetric excited states reachable, which the merged DALTON-seeding work (#755) benefits from. A review when someone has a moment would be much appreciated.

ahurta92:

PR #757, opened 2026-08-17
molresponse: one engine for the quadratic sources (V^BC and the 2PA residue as specs)

ahurta92:
## What this adds

The hyperpolarizability source (`compute_vbc`) and the two-photon residue source turn out to share a three-part structure, differing only in which orbital pairs, signs, and transposes go where. This PR adds a small declarative engine that assembles a source from that data:

- `kernels/source_spec.hpp` — a `SourceEntry` describes one term ({pairs, density, target, sign, apply-vs-matrix mode, transpose, projection}); `assemble_source` evaluates a list of them with the existing primitives (`apply_gamma_raw`, `matrix_inner`, `transform`). No new physics.
- `compute_vbc_spec` — V^BC expressed as a spec, added **beside** `compute_vbc` (which is unchanged).
- `kernels/tpa_source_spec.hpp` — the 2PA residue right-hand side `(P,Q)` as the second spec. This gives the state-independent form: build `(P,Q)` once from the two photon-leg responses, then the moment is one inner product with the excited-state vector.

Production paths are untouched — both new specs sit alongside the existing builders, with equality tests wired in as ctest gates.

## Why

Follow-up properties (excited-state-to-excited-state moments, resonant Raman) need sources of the same shape; with this they become data, not new builders. It also gives one place for the symmetrization/projection/truncation policy that's currently duplicated. The design mirrors how DALTON structures the same machinery (one E[3] builder serving its whole property family).

## Checks

- `compute_vbc_spec` vs `compute_vbc`: per-channel difference ~3e-13 on both stand-ins and solved responses (tolerance was 1e-10); β through both paths identical to ~9e-14; the Kleinman pair asymmetry comes out identical through both paths.
- `(P,Q)` spec vs the production residue composition: 2e-8 relative.
- Full response ctest suite: 19/19 (including the two new gates).

Naming and the entry-struct shape are open to suggestions — this is meant to be the foundation other properties build on, so better ideas now are cheap.


ahurta92:

PR #758, opened 2026-08-17
Fix: Inconsistency in `mul()` and `dot()`

EricaCMitchell:
## Description

Add `mul_sparse()`/`dot_sparse()` for `std::vector<Function>`, route the exchange operator's screened multiplies through them, and remove the duplicate `tol` parameter from `mul()`/`dot()`. Before this change a caller could screen a vector multiply two ways — `mul(..., tol=...)` or the scalar `mul_sparse()` inside a loop — and only one of those two paths was correct. Now there is one way to screen: `mul_sparse()`/`dot_sparse()`, with `tol` as an explicit, un-defaulted parameter placed to match the scalar functions' order.

## Details

- [X] **Add `mul_sparse()` and `dot_sparse()` for vectors** — [d54459d98](https://github.com/m-a-d-n-e-s-s/madness/commit/d54459d9828b5c12d06659c5c0954ed89e75a14e). New vector overloads of `mul_sparse()`/`dot_sparse()` in `vmra.h`, mirroring the shape of the existing `mul()`/`dot()` vector functions: `mul_sparse()` makes both operand vectors redundant once (not once per pair), then calls the scalar `mul_sparse(a[i], b[i], tol, false, false)` per pair; `dot_sparse()` calls `mul_sparse()` and sums. `tol` is a required parameter, not a default, so a caller cannot silently get the unscreened path.
- [X] **Use `dot_sparse()` in the exchange operator** — [6a96f2ce3](https://github.com/m-a-d-n-e-s-s/madness/commit/6a96f2ce3c792d2bacec3033c17b235e97f0dea2). Both screened multiplies in `exchangeoperator.cc`/`.h` (`MacroTaskExchangeSimple::compute_offdiagonal_batch` and the diagonal path) called `dot(world, ..., tol=mul_tol)` — the `tol` parameter that this PR goes on to remove. Switched to `dot_sparse(world, ..., mul_tol, true, true)`, the function this parameter was meant to route through.
- [X] **Rename the `dot()` test to `dot_sparse()`** — [4012e6a9c](https://github.com/m-a-d-n-e-s-s/madness/commit/4012e6a9c12c66e29a899bd424cf53c3bd2094ce). `test_mul_sparse.cc`'s T5 called `dot(world, a, b, true, true, tol)`. Renamed the test function (`test_dot` → `test_dot_sparse`) and its print/checkpoint messages, and switched the call to `dot_sparse(world, a, b, tol, true, true)`, so T5 now exercises the same entry point as the exchange operator.
- [X] **Remove the `tol` parameter from `mul()` and `dot()`** — [3b074ea28](https://github.com/m-a-d-n-e-s-s/madness/commit/3b074ea286edb32f49bb14fa171fcfc3d60034ff). Drops the `double tol=0.0` parameter from the scalar `mul()` (`mra.h`) and the vector `mul()`/`dot()` (`vmra.h`); the scalar `mul()` now always calls `mul_sparse(left, right, /*tol=*/0.0, fence, do_make_redundant)`. This removes the duplicate way to set the screening level and gives `mul()`/`dot()` the same parameter order as `mul_sparse()`/`dot_sparse()`. Any remaining in-tree caller passing a nonzero `tol` to `mul()`/`dot()` needs to move to `mul_sparse()`/`dot_sparse()` — the two call sites in the exchange operator already were, in the preceding commit.

## AI

- Commit drafting and PR message assembly were AI-assisted; the code changes and verification were written and run by the author.

### DIFF
--- a/src/madness/mra/vmra.h
+++ b/src/madness/mra/vmra.h
@@ -1205,6 +1205,49 @@ namespace madness {
         return vmulXX(a, v, tol, fence);
     }
 
+    /// Multiplies two vectors of functions using sparsity of a[i] and b[i] --- q[i] = a[i] * b[i]
+    ///
+    /// Box pairs whose estimated contribution falls below the tolerance are skipped instead
+    /// of being multiplied. Both inputs are made redundant; the screening reads their
+    /// norm_tree and dnorm_tree.
+    ///
+    /// Leaves both inputs in redundant form. Function is a shallow handle, so this is visible
+    /// to the caller: logically const, not bitwise const. Converting back is not free, so a
+    /// caller that reuses the operands afterwards must do it itself.
+    ///
+    /// @param[in] tol  target absolute accuracy of the product; the safety margin is applied
+    ///                 internally (FunctionImpl::MUL_SCREENING_SAFETY), so pass the accuracy
+    ///                 wanted, not a pre-scaled value. tol=0 multiplies exactly. The criterion
+    ///                 estimates the neglected cross terms rather than bounding them: the error
+    ///                 tracks tol up to a measured O(1-20) constant and decays as ~tol^0.75
+    ///                 rather than ~tol (see test_mul_sparse.cc). The meaning differs from the
+    ///                 earlier norm_tree-based screen, so a previously tuned value needs
+    ///                 re-checking.
+    /// @param[in] do_make_redundant  if false, both inputs must already be redundant
+    template <typename T, typename R, std::size_t NDIM>
+    std::vector< Function<TENSOR_RESULT_TYPE(T,R), NDIM> >
+    mul_sparse(World& world,
+        const std::vector< Function<T,NDIM> >& a,
+        const std::vector< Function<R,NDIM> >& b,
+        double tol,
+        bool fence=true,
+        bool do_make_redundant=true) {
+        PROFILE_BLOCK(Vmulvv);
+        if (do_make_redundant) {
+            // prepare once, not once per pair: mul_sparse fences whenever it prepares.
+            // Redundant inputs make the second call a no-op, so no aliasing check is needed.
+            make_redundant(world, a, false);
+            make_redundant(world, b, false);
+            world.gop.fence();
+        }
+        std::vector< Function<TENSOR_RESULT_TYPE(T,R),NDIM> > q(a.size());
+        for (unsigned int i=0; i<a.size(); ++i) {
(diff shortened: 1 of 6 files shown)


hborchert (approved):

src/madness/mra/vmra.h
```diff
@@ -1204,6 +1204,49 @@ namespace madness {
...
+    template <typename T, typename R, std::size_t NDIM>
+    std::vector< Function<TENSOR_RESULT_TYPE(T,R), NDIM> >
+    mul_sparse(World& world,
+        const std::vector< Function<T,NDIM> >& a,
+        const std::vector< Function<R,NDIM> >& b,
+        double tol,
+        bool fence=true,
+        bool do_make_redundant=true) {
+        PROFILE_BLOCK(Vmulvv);
+        if (do_make_redundant) {
```
fbischoff, line 1242: these lines will force a fence, even if it has been set to false explicitly. Maybe you would want a check on that. Alternatively use enforce_tree_state, which throws a runtime error in that case

EricaCMitchell: I'm looking into this. I will also change other sites that have this same issue and replace the non-respected fence with `ensure_tree_state_respecting_fence`

EricaCMitchell: Fixed, I've changed the call sites to use `ensure_tree_state_respecting_fence`. [b16f28d](https://github.com/m-a-d-n-e-s-s/madness/pull/758/commits/b16f28d8a00e65eff7b36a81cfa789210f36bb31) [cdb71dc](https://github.com/m-a-d-n-e-s-s/madness/pull/758/commits/cdb71dc26ad70929d658841306fc18cb958f37b1)

It does something similar to [gaxpy](https://github.com/m-a-d-n-e-s-s/madness/blob/16ec3cd5a8e9169c46f4e9844a7f7ed6d492f9ab/src/madness/mra/vmra.h#L1859-L1866) where if it's possible to follow the requested fence, it will. Otherwise, it will notify the user the fence could not be respected. This was chosen because if the user explicitly requests `do_make_redundant == true`, then it would throw an error if the fence could not be respected, which would be no different than `do_make_redundant == false`.

In short, the user's choice of `do_make_redundant` supersedes the forced fence. If they truly want to avoid a hidden fence, then they'll need to provide functions in the redundant state before the `mul` or `mul_sparse` calls.

EricaCMitchell:
@fbischoff Any other reviews from the last two commits?

Otherwise, I'll merge to master this weekend.

PR #761, opened 2026-08-18
molresponse: root-identity guard for seeded excited-state solves

ahurta92:
## What this adds

When an excited-state solve starts from a DALTON seed, it refines the states it was handed — it doesn't search for the N lowest. That's often what you want (state tracking), but it has a quiet failure mode: if the seeding basis can't describe a low-lying state (e.g. a diffuse state in an unaugmented basis), the seeded solve converges quickly, reports converged, and silently substitutes a different state.

This PR makes that visible:

- after a seeded solve, each root gets {overlap with its seed, seed ω, ω shift} recorded in the metadata and printed as a summary table
- a note prints when the solve was pure tracking (every root overlap > 0.9, converged in ≤ n_roots iterations), saying plainly that N-lowest is not guaranteed
- a loud warning fires if a converged root has low overlap with every seed root (basin escape)
- an optional cross-check, `--es-expect-omegas=w0,w1,...` (deck: `excited.expect_omegas`), warns for any root farther than a tolerance (default 0.02 au) from every expected value — the intended use is handing a well-augmented reference ladder to a run seeded from a smaller basis

## The case that motivated it

H2O seeded from cc-pVDZ: root 3 converges in 2 iterations, converged=true, at 0.4626 au — but the true 4th-lowest state is at 0.4096 (all augmented seeds find it). The cc-pVDZ basis simply can't describe the diffuse state, so its CIS root 3 *is* a different state, and the solve faithfully refined it. With this PR, that run prints the tracking note and (given the reference ladder) a warning naming the substituted root; the same run seeded from d-aug-cc-pVDZ prints a clean all-roots-OK line.

Both behaviors are reproduced in the validation runs (transcripts in the PR branch history). Response ctest suite passes; a docs paragraph states the tracking semantics.


ahurta92:

PR #770, opened 2026-08-21
Add `#include <memory>` to `periodic/test.cc`

JonathonMisiewicz:
Fix a missing include leading to this test failing to compile on Linux. Smart pointers require the memory header.

### DIFF
--- a/src/examples/periodic/test.cc
+++ b/src/examples/periodic/test.cc
@@ -1,5 +1,6 @@
 #include "testpc.h"
 #include <fftw3.h>
+#include <memory>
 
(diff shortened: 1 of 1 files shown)


devreal (changes requested):
Thanks @JonathonMisiewicz! On quick glance the following are also missing:

- `<vector>`
- `<algorithm>` (for `std::reduce`)
- `<iostream>` (for `std::cout`)

JonathonMisiewicz:
Good catch - added those as well.

devreal (approved):

PR #772, opened 2026-08-22
Pr fix open shell gga

fbischoff:
fixes open-shell GGA bug


fbischoff:
commits in here are duplicated in pr-implement-metaGGA

PR #778, opened 2026-08-25
Improve Probe Displacement Selection for All-Even Case

JonathonMisiewicz:
Supercedes #732. The new algorithm is much simpler.

1. Displace on the origin of the smallest odd-radius face if possible.
2. If we can't do that, choose the smallest even-radius face. That's equivalent to the origin, so now we displace a half-radius along another dimension. Choose the second-smallest even radius face. Failing that, just choose the first open face.

Tests pass locally. Testing on cluster in-progress. Don't merge until that's done.

Changes are expected all across the board, but the even ones are the most important.


evaleev (approved):

JonathonMisiewicz:
Good to merge whenever you want.

JonathonMisiewicz:
MPQC integration testing of this PR blocked until #786 comes in.

evaleev (approved):

PR #780, opened 2026-08-26
Pr bsh apply

hborchert:
# PR title: BSH apply backends with a single-owner operand redistribute

## What this does

Makes the `compute_residual` BSH (Helmholtz) apply parameter-selected, with a macrotask backend
that localizes each operand function onto one rank and applies rank-locally. Default `auto`:
macrotask when multinode or at tight threshold, tile on a single node at loose/medium. The
previous hardwired tile loop (tile size independent of the rank count) and a dead inline
`ApplyTask` are replaced.

**1. `world/worlddc.h` — single-owner pmap + coalesced redistribute.**

- `WorldDCSingleOwnerPmap`: every key on one fixed rank; `single_owner()` on the pmap
  interface returns that rank, −1 for any other pmap
- `redistribute_coalesced_phase1/2`: one bulk `insert_batch` AM per (destination, chunk)
  instead of one AM per box; erase-after-copy, so the move streams instead of doubling
- the caller owns the fences; the one between the phases is required (phase1 iterates the
  ConcurrentHashMap)

**2. `mra/funcimpl.h` — cross-world copy fast paths.**

- single-owner pmap: fetch from that rank only, not an all-ranks poll
- own-rank shard: copy nodes directly, no serialize round-trip; the existing all-ranks poll
  takes the same shortcut for its own-rank turn

**3. `mra/vmra.h` — function-level redistribute.**

- `redistribute_to_batches(world, v, owner)`: v[j]'s whole tree moves to rank owner[j];
  state-preserving; owner[] agreement checked collectively
- `assign_round_robin`, `assign_cost_aware` (deterministic LPT), `function_costs`
  (replicated per-function cost proxy, one reduction)

**4. `chem/SCF.*` — the executors and the dispatch.**

- `apply_bsh_macrotask` (rank-local apply in subworlds, in-subworld truncate),
  `apply_bsh_tiled` (rank-aware tile size), `apply_bsh_plain`
- at tight threshold (≤ 1e-7) the operand is pre-localized cost-aware and each task pinned to
  its operand's rank, so the framework's input copy is a local fetch; an explicit conflicting
  `bsh_apply_nworld`/batch setting wins and skips the redistribute with a notice
- the residual `transform` is tiled over output columns: untiled it materializes all nmo fpsi
  next to psi and Vpsi (~3x orbital memory at tight thresh), the OOM site for large systems

## Commits

| commit | what |
|---|---|
| `556e2991a` | add WorldDCSingleOwnerPmap and a pmap single_owner query |
| `225230dea` | add single-owner and same-rank fast paths to the cross-world coeff copy |
| `1ace40f06` | coalesce the container redistribute into bulk per-destination messages |
| `e2e9785b6` | redistribute whole functions onto single owner ranks |
| `635654010` | select the compute_residual BSH apply by parameter, tile or macrotask |
| `4ed6ba434` | tile the residual transform over output columns |
| `aabb775c3` | pre-localize the macrotask BSH operand at tight protocol |

7 commits, +662/−77 over 8 files.

## Interface changes

- new parameters: `bsh_apply` (`auto`|`tile`|`macrotask`|`plain`, default `auto`),
  `bsh_apply_nworld`, `bsh_apply_{min,max}_batch`, `bsh_apply_max_tile`, `bsh_apply_policy`,
  `bsh_apply_memcheck`. Defaults reproduce the previous energies to the cross-run noise floor
- `print_level >= 2` logs the auto decision and the redistribute; `>= 10` adds per-task
  apply/truncate, cloud copy/read times and peak RSS
- no checkpoint-format change

## PR checklist

| item | answer |
|---|---|
| raw MPI calls | none added |
| blocking MPI collectives | none added; the redistribute uses AMs + `World::gop` fences |
| implicit default World | none added |
| non-collective container construction | none. phase1/phase2 are called on every rank between collective fences; the owner assignment is replicated and checked collectively |
| `Function` mutation from user threads | no. The redistribute runs between fences from the caller; kernels run in tasks |
| tree-state preconditions | unchanged; `redistribute_to_batches` is state-agnostic (tested reconstructed and compressed) |
| GENTENSOR coverage | test_dc, testvmra, test_cloud, test_vectormacrotask pass at np=1/2 in a Debug build with assertions and `-DENABLE_GENTENSOR=ON` (Pthreads backend); testsuite passes at np=1/2 |

### DIFF
--- a/src/madness/chem/CalculationParameters.h
+++ b/src/madness/chem/CalculationParameters.h
@@ -68,10 +68,10 @@ struct CalculationParameters : public QCCalculationParametersBase {
 		initialize<double>("charge",0.0,"total molecular charge");
 		initialize<std::string> ("xc","hf","XC input line");
 		initialize<std::string> ("hfexalg","multiworld_row","hf exchange algorithm; multiworld bounds memory, needs one subworld per rank",{"multiworld","multiworld_row","fetch_compute","smallmem","largemem"});
-		initialize<long>  ("hfex_batch_granularity",1,"hfexalg=multiworld: orbital batches per rank; >1 balances better");
-		initialize<bool>  ("hfex_cost_aware_assign",true,"hfexalg=multiworld: place tasks by measured cost, not by count");
-		initialize<int>   ("hfex_local_accumulation",2,"hfexalg=multiworld: gather tile results 1=per subworld, 2=per node");
-		initialize<std::string>("bsh_apply","auto","BSH operator apply backend in compute_residual: auto=macrotask multinode or at tight thresh, tile otherwise; tile=rank-aware memory-bounded loop; macrotask=distribute across subworlds (rank-local apply, memory-frugal); plain=single un-tiled apply",{"auto","tile","macrotask","plain"});
+		initialize<long>  ("hfex_granularity",1,"hfexalg=multiworld: orbital batches per rank; >1 balances better");
+		initialize<bool>  ("hfex_cost_aware",true,"hfexalg=multiworld: place tasks by measured cost, not by count");
+		initialize<int>   ("hfex_accumulation",2,"hfexalg=multiworld: gather tile results 1=per subworld, 2=per node");
+		initialize<std::string>("bsh_apply","auto","BSH apply backend; auto=macrotask when multinode or at tight thresh, else tile",{"auto","tile","macrotask","plain"});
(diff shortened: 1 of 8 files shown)


fbischoff (dismissed):
looks good to me.

If you still have a couple of hours to spare for testing you could implement the changes also in BSHApply which is used in nemo and some other places. Or you could use BSHApply also in SCF. Since the basic structure is always the same (new = G(V*old - coupling(old) ) I've factored that out for reuse.

If that is too much changes for now I can approve anyways, just let me know

fbischoff (approved):

PR #784, opened 2026-09-01
Pr localizer

hborchert:
# Distributed orbital localization: localize=new_sys, localize=cholesky

`localize new` runs its optimizer on rank 0 only; beyond ~1500 orbitals a
single call exceeds the idle-rank watchdog (`MAD_WAIT_TIMEOUT`) and aborts
the job. This PR adds two distributed localizers that are fast and
watchdog-safe at that scale, plus small SCF-side localization changes. The
existing optimizers (`new`, `boys`, `pm`) are untouched.

- `localize new_sys`: the "new" objective (Pipek–Mezey in the orthonormal
  atomic-eigenfunction basis) optimized with the distributed systolic
  Jacobi-sweep engine of `distributed_localize_PM`. Localize-set boundaries
  are respected during the sweeps (in moldft, multiple sets are populated
  only for `pm`; the nemo stack enforces core–valence separation through
  the same interface).
- `localize cholesky`: non-iterative, deterministic localization by pivoted
  Cholesky of the density in the orthonormal atomic-eigenfunction basis
  (Aquilante, Pedersen, Koch 2006). The pivot search is divided over ranks.
  Each SCF iteration reuses the previous iteration's pivot order; a slot
  re-pivots only when a candidate beats the remembered pivot by 2x.
- The orbital transform is skipped when the computed rotation is
  near-identity (max off-diagonal <= 0.01), keeping the KAIN history in one
  orbital basis across skipped iterations.
- The loosened (100x) localization tolerance on the initial-guess iteration
  is removed.
- The SCF localize timer is split into matrix and transform phases
  (`print_level >= 3`).

All localizers give results independent of rank count. No new input
parameters beyond the two `localize` values. No serialization or
checkpoint-format change.

## Commits

| commit | what |
|---|---|
| `d07e403c8` | `localize=new_sys`: the new localizer's objective via distributed systolic sweeps |
| `a6128adba` | split the localize timer into matrix and transform phases |
| `47eda78a1` | `localize=cholesky`: non-iterative localization by pivoted Cholesky |
| `5d6ca7638` | remove the loosened localization tolerance on the initial-guess iteration |
| `cc9d2439d` | cholesky pivot memory + skip near-identity localization rotations |
| `76640a95b` | clarify the PM_q comment: an empty S signals an orthonormal basis (review) |
| `26f2c9bb2` | localize phase timing: sum matrix/transform over both spin blocks (review) |

7 commits, +334/−32 over 7 files.

## Semantics

- `new` and `new_sys` are different optimizers of the same objective.
  Converged SCF energies agree (valinomycin, full protocol to k10:
  0.07 uHa; identical iteration counts).
- `new` and `boys` are unchanged. The only edit near them factors the
  "new" basis preparation into a shared helper (`prepare_new_basis`, used
  by `new_sys` and `cholesky`); `new` reproduces the pre-PR result
  digit-for-digit (water5).

## Tests

No new unit tests; `test_localizer` is unchanged. Application-level checks:

- water5: `new` matches the pre-PR binary digit-for-digit; `new_sys`
  bitwise identical at 1 vs 8 ranks; multi-set (core/valence) clean.
- valinomycin k10: `new_sys` / `cholesky` converge to the `new` energy
  within 0.07 / 0.05 uHa.
- Pivot memory, input-identical A/B (1546-MO protein): KAIN
  subspace warnings 16 -> 0, monotone energy descent.
- Timing, warm localize call at 1535 MOs / 24 ranks: 2172 s (`new`, rank 0)
  -> 88 s (`new_sys`).

## PR checklist

| item | answer |
|---|---|
| raw MPI calls | none |
| blocking MPI collectives | none; reductions via `world.gop` |
| implicit default World | none |
| non-collective container construction | the added localizers run collectively on every rank; `new`/`boys` keep their rank-0 + broadcast structure |
| Function mutation / tree-state preconditions | rotations via the unchanged `transform` machinery |
| GENTENSOR coverage | no tensor-layer change |

## Verification

- CI green (12/12) at `d4cc00554`; rerunning at `26f2c9bb2` (review edits
  and scope reduction only).
- In production since 2026-08: the `new_sys` and `cholesky` paths ran the
  full Rudberg polypeptide benchmark set (150-790 orbitals) and a 1065-MO
  localized-SCF campaign.

### DIFF
--- a/src/madness/chem/distpm.cc
+++ b/src/madness/chem/distpm.cc
@@ -72,7 +72,10 @@ void matrix_inner(DistributedMatrix<T>& A,
 static inline double PM_q(const tensorT & S, const double * MADNESS_RESTRICT Ci, const double * MADNESS_RESTRICT Cj, int lo, int nbf)
 {
     double qij = 0.0;
-    if (nbf == 1) { // H atom in STO-3G ... often lots of these!
+    if (S.size() == 0) { // empty S marks an orthonormal block (S == identity): the "new" method
+        for(int mu = 0;mu < nbf;++mu) qij += Ci[mu + lo] * Cj[mu + lo];
+    }
+    else if (nbf == 1) { // H atom in STO-3G ... often lots of these!
         qij = Ci[lo]*S(0,0)*Cj[lo];
     }
     else {
@@ -386,6 +389,42 @@ DistributedMatrix<double> distributed_localize_PM(World & world,
     // }
     // world.gop.broadcast(U.ptr(), U.size(), 0);
     //return U;
+}
+
+DistributedMatrix<double> distributed_localize_new(World& world,
+                                                   const Tensor<double>& C,
+                                                   const std::vector<int>& set,
+                                                   const std::vector<int>& at_to_bf,
+                                                   const std::vector<int>& at_nbf,
+                                                   const double thresh,
+                                                   const double thetamax)
+{
+    const long nmo = C.dim(0);
+    const long nao = C.dim(1);
+    const long natom = at_to_bf.size();
+
+    // Empty overlap blocks select PM_q's orthonormal fast path (S == identity): in the
+    // orthonormal atomic-eigenfunction basis the charge q_ij(a) is a plain dot product.
+    std::vector<tensorT> Svec(natom);
+
+    DistributedMatrix<double> dU = column_distributed_matrix<double>(world, nmo, nmo);
+    dU.fill_identity();
+
+    DistributedMatrix<double> dC = column_distributed_matrix<double>(world, nmo, nao);
+    dC.copy_from_replicated(C);
+
+    DistributedMatrix<double> dA = concatenate_rows(dC, dU);
+
+    world.taskq.add(new SystolicPMOrbitalLocalize(dA, set, at_to_bf, at_nbf, Svec, thresh,
+                                                  thetamax, natom, nao, nmo));
+    world.taskq.fence();
+
+    dA.extract_columns(nao, nmo+nao-1, dU);
+
+    world.taskq.add(new SystolicFixOrbitalOrders(dU));
+    world.taskq.fence();
+
+    return dU;
 }
 
 }
--- a/src/madness/chem/distpm.cc
+++ b/src/madness/chem/distpm.cc
@@ -72,7 +72,7 @@ void matrix_inner(DistributedMatrix<T>& A,
 static inline double PM_q(const tensorT & S, const double * MADNESS_RESTRICT Ci, const double * MADNESS_RESTRICT Cj, int lo, int nbf)
 {
     double qij = 0.0;
-    if (S.size() == 0) { // empty S marks an orthonormal block (S == identity): the "new" method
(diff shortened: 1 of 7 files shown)


fbischoff (commented):
as far as I understand, "new" and "new_sys" are the same from a user's perspective? I'd rather have one of the two default and the other one removed, unless there is an advantage of keeping the old one.

src/madness/chem/distpm.cc
```diff
@@ -72,7 +72,10 @@ void matrix_inner(DistributedMatrix<T>& A,
 static inline double PM_q(const tensorT & S, const double * MADNESS_RESTRICT Ci, const double * MADNESS_RESTRICT Cj, int lo, int nbf)
 {
     double qij = 0.0;
-    if (nbf == 1) { // H atom in STO-3G ... often lots of these!
+    if (S.size() == 0) { // empty S marks an orthonormal block (S == identity): the "new" method
```
fbischoff, line 75: This comment doesn't make sense: if S.size()==0 it's non-existent, not orthonormal

robertjharrison:
The "new" localizer should not by default have its orbitals divided into core-valence sets --- by localizing into the atomic eigenfunctions it should achieve stable energy separation much more effectively on its own, and especially on heavier elements where it will distinguish the n=1,2,.. shells.  We had to do it for the other localizers because they would routinely screw up the core especially for extended or heavy systems.   Per previous conversation with Hannes, the virtual orbitals (or any non-unit occupation) need to be resolved by diagonalizing the relevant block of the fock matrix (before BSH orbital update is applied, i.e., at the same time that localization is done or at least once you have the fock matrix, not forgetting to transform that before using it in the update).

Can I suggest that use of core-valence sets is made optional with the default being off, and we can in the documentation advise people to turn it on if convergence issues are encountered esp. with other localizers.  Or should we make the default dependent on which localizer is chosen?

Is there are recommendation to prefer new_sys or new_rank_parallel implementations?  They do seem redundant, though I note that one does Jacobi rotations and the other does CG which could have very different convergence characteristics.  Do tests reveal any performance or localization result differences?

Is some variant of the new localizer still the default?

[Aside: after Hannes determined the issue I have fixed/reviewed all of the multithreaded task code and stress tested it on both arm and x86, with the minimal fixes for arm pushed to the fast linear algebra PR.  Amended to note that Ed has further refined the solution in a PR to a PR that is currently approved but not yet merged.]

robertjharrison:
Can we pls close this out? Need to break the log jam.

hborchert:
The localizer dependent default is already the case, core-valence sets are only used by PM and are otherwise turned off by default. New_sys reuses the same systolic engine as PM which respects a set vector when handed one. 

The default localizer is still 'new' with CG and tests for small up to very large MO counts show no convergence-quality difference. Systolic is a lot faster compared to CG for the 1000+ MO systems, hence 'new_sys' is the preferred option for very large systems but not needed for system sizes up to 300 MOs. I only tested a couple of molecules and kept CG the default since Jacobi could converge differently for a given system.

robertjharrison (approved):
Other than needing to reconsider naming convention merge now to facilitate full testing and release.

PR #783, opened 2026-08-31
molresponse: optional fine-rung start for DALTON-seeded response solves

ahurta92:
Small follow-up to #755. When a response solve starts from a DALTON seed, the protocol ladder's coarse rung can erase the seed's head start — at frequencies between poles we measured the coarse rung hitting the iteration cap in every run (cold or seeded), so both entered the fine rung from the same state and the seed bought nothing.

This adds `seed.start_rung fine` (deck) / `--seed-start-rung=fine` (driver): with a seed present, the ladder collapses to its finest rung and the seed projects directly at that resolution. Default is unchanged (`coarse` = today's behavior); without a seed the option warns and does nothing.

Unit tests for the ladder logic are included and pass; an iteration-count A/B at the between-pole points is queued and will be posted here when it lands. Marked draft until then.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Vn5VcqgPWNcd9pEcTWjj5d

PR #785, opened 2026-09-02
molresponse: consolidated rollup — ES guess UI, seeded-ES root guard, quadratic source engine, seed fine-rung

ahurta92:
This consolidates the four open molresponse PRs (#754, #757, #761, #783) into one branch so there is a single source to review and merge against. Each piece keeps its own commits, so it can still be reviewed commit-by-commit; the individual PR threads hold the discussion and evidence so far and are closed with pointers here.

The branch is upstream master + the four feature branches merged in train order. One deliberate union resolution in `src/madness/chem/ResponseParameters.hpp` (both #754's `excited.guess`/`excited.guess_basis` and #761's `excited.expect_omegas`/`expect_tol` are kept).

### 1. ES guess UI (was #754) — `becc88717`
Exposes the existing `create_virtual_ao_guess` (NWChem CIS-diagonal guess) as `--es-guess` / deck `excited.guess`, with `--es-guess-basis` / `excited.guess_basis`. Motivation: the default solid-harmonics trials are purely angular, so totally-symmetric states on atoms (e.g. He 2^1S) are unreachable at any root count. Also makes the silent short-guess pad-with-solid-harmonics top-up loud. Defaults unchanged (solid_harmonics / aug-cc-pvdz); existing decks and CLIs behave identically.

Known follow-up (not in this rollup): `--es-roots` currently plans the ES bundle plus derived dipole FD legs at half of each root energy (the two-photon ingredients) even when no 2PA is requested. A follow-up will make roots-only the default and gate the FD legs + contraction behind an explicit `--tpa` / `excited.tpa`.

### 2. Root-identity guard for seeded ES solves (was #761) — `f382cb813`, `231c1c57b`
A seeded excited-state solve refines the states it is handed; it does not search for the N lowest. Observed: a cc-pVDZ-seeded solve converged (2 iterations, converged=True) to 0.4626 au where the true root is 0.4096. The guard adds per-root seed-overlap and omega-shift reporting plus `--es-expect-omegas` / `excited.expect_omegas` cross-checks that warn when a converged root sits more than `expect_tol` from every expected value. Reporting only — no solver behavior change.

### 3. Quadratic source engine (was #757) — 6 commits
One declarative evaluator (`SourceEntry` specs) for the quadratic sources: V^BC (hyperpolarizability) and the state-free 2PA residue (P,Q) become two data tables over the same engine. Commit order tells the story: two pinning tests first (regroup identities M1/M2/M3, then (p,q) vs V^bc), then the engine, then V^BC as a spec beside the existing path, then the residue promoted to kernels/ as a spec, then the ctest wiring. Equality gates against the legacy paths: 3e-13 / 9e-14 / 2e-8.

### 4. Seed fine-rung start (was #783) — `27d041709`
`seed.start_rung fine`: DALTON-seeded FD runs can skip the coarse protocol rung. Between poles the coarse rung hits maxiter unconverged in every arm and erases the seed's head start. Measured A/B (H2O / d-aug-cc-pVTZ, three between-pole points, details in #783): full ladder cold = 3x60 discarded coarse + 18/27/37 fine; cold fine-start = 29/33/41; seeded fine-start = 8/22/25 — same alpha_zz to ~4e-4 relative across all arms. Below-pole regression: default path unchanged (k6=9, k8=4 vs baseline); seeded fine-start 5 iters vs 13. Default off; opt-in flag.

## Validation

- Each piece's branch passed CI individually on its own PR.
- Local witness on this exact merge (xeonmax, Release, C++20): build + the 14 molresponse ctests — result will be posted in a comment when the job lands.
- Evidence for the science claims: #754 (He 2^1S reachability), #761 (root-tracking observation), #757 (equality gates in the two pinned tests), #783 (iteration A/B tables on the thread).

https://claude.ai/code/session_01KtdsZQ67mYUjLqj3zfwepk


ahurta92:
Local build + test witness for this exact merge (`3bd725ad4`), as promised in the description:

- Platform: Stony Brook Seawulf xeonmax (Sapphire Rapids HBM), gcc 13.2, Debug, TBB backend, C++20, fresh build dir.
- Build: clean (`BUILD_RC=0`), 165 targets including all molresponse test binaries.
- Tests: **16/16 molresponse ctests pass** — the 14 pre-existing suite tests plus the two new pinning tests this rollup adds (`test_tpa_regroup_identities`, `test_tpa_pq_vs_vbc`). Two scripted madqc E2E python tests are skipped at this tier (they need the full madqc build).

Jobs 2145921/2145932; ctest transcript in the build dir's `Testing/Temporary/LastTest.log`.

ahurta92:
Walkthrough review of the pieces is underway (guess UI, root guard, and the source engine done so far); per our plan, findings land on this branch as they are found. Three commits pushed, witnessed together (xeonmax Debug/TBB: build clean, 16/16 molresponse ctests):

**`fc04af5d4` — ES requests are roots-only by default; `--tpa` / `excited.tpa` opts into the two-photon ingredients.** Review of the guess-UI piece showed the Resonant plan unconditionally emitted 3 derived dipole FD legs per root at omega_n/2 (the 2PA machinery, roughly 5x the ES-leg cost) even when no 2PA was requested — and on the standalone/madqc surfaces the 2PA contraction was unreachable anyway, so users paid for ingredients with no way to ask for the number. Now: `ResponsePropertyRequest.tpa` (default false) gates the legs; `--tpa` (standalone + test driver) and `excited.tpa` (deck) set it; the shared workflow runs `assemble_tpa` whenever the plan carries the legs, so all three surfaces can actually produce the 2PA number. Restart chaining is the intended path: a roots-only run followed by a `--tpa` run in the same calc dir picks the converged ES bundle up from metadata and solves only the legs. Also unifies the `es_freq_factor` default (was 1.0 in CalcNode vs 0.5 in the request — always overwritten today, but a landmine for a hand-built node).

**`c1c83dd8d` / `087f07812` — the two scheduler-test casualties of that change.** The resonant-DAG case pins the derived-FD node mechanics, so it now requests tpa; the schedule-exclusion case was passing trivially once the roots-only plan carried no symbolic nodes, so it does too. `087f07812` also adds a scope note to `kernels/source_spec.hpp` from the source-engine review: its g' is J − c_xc·K (HF exchange only), and `eval_fock_action` is one of the sites a future DFT f_xc extension must cover — stated in the header so it can't be missed.

One behavior change to be aware of when reviewing: a deck with `excited.enable` (and no `excited.tpa`) now solves the ES bundle only — previously it also ran the derived FD legs it had no way to use.

ahurta92:
Self-review summary — each piece of this rollup has now been walked through line-by-line. Recording the outcomes here so reviewers know what was already looked at and what came out of it.

**1. ES guess UI** (`becc88717`) — reviewed, one design issue found and fixed on this branch. The `--es-roots` path always planned the derived dipole FD legs at omega_n/2 (two-photon ingredients, ~5x the ES cost) even when no 2PA was requested, and two of the three user surfaces had no way to run the 2PA contraction anyway. Fixed in `fc04af5d4`: ES requests are roots-only by default; `--tpa` / deck `excited.tpa` opts into the FD legs plus the contraction, on all three surfaces. Note the behavior change: a deck with `excited.enable` alone now solves the ES bundle only.

**2. Root-identity guard** (`f382cb813`, `231c1c57b`) — reviewed, clean, no changes. Two non-blocking notes on record for future work: (a) the guard deep-copies the seed bundle and holds it in memory for the whole solve — fine for small systems, a real footprint at ~30+ occupied orbitals; reduced-k copies would suffice for the 0.5/0.9 overlap thresholds. (b) A matched seed root missing its ladder entry prints `omega_seed 0.000000` instead of `-` (cosmetic, near-impossible edge).

**3. Quadratic source engine** (6 commits) — reviewed, clean. The "engine contains no physics" claim checks out: the evaluator only sequences existing primitives, and the V^BC table is a term-for-term image of `compute_vbc_i`. The validation ladder (identities measured on real MRA functions -> spec-vs-bespoke equality at 1e-10 with four orders of separation from the failure scale -> old path kept alive beside the new one) is enforced in ctest, not a one-time claim. One addition from review (`087f07812`): a scope note in the header stating g' is HF-exchange-only, so the future DFT f_xc extension cannot miss `eval_fock_action`.

**4. Seed fine-rung start** (`27d041709`) — reviewed, no changes needed. Default is a verified no-op (below-pole regression matches the old baseline exactly), misuse warns or errors loudly, edge cases are unit-tested, and the between-pole iteration A/B (posted on #783) measures the win directly.

Validation state: all 12 CI lanes green on the current head; local witness (Seawulf xeonmax, Debug/TBB, C++20) builds clean with 16/16 molresponse ctests passing.

ahurta92:
One more review-driven commit (`73d2fb317`): an audit of molresponse against master's recent changes (restartdata v5, exchange screening re-enabled in #724, the mul_sparse tol-semantics change, C++20, the misc/mra header moves) found the code consistent on all counts except one — the post-convergence transition-moment analysis passed `vtol = thresh` to mul_sparse where every other site (and the SCF.cc analysis block this code mirrors) uses `thresh*0.1`. Aligned. Diagnostic-only path; the audit details live in the workspace notes. Worth knowing: the 16-test tier passing on this branch already validates molresponse under the re-enabled exchange screening and new mul_sparse semantics — including the two 1e-10-gated source-equality tests.

ahurta92:
Correction to my earlier witness comment, plus new validation work — recorded here so reviewers have the accurate picture.

**Correction.** The earlier local-witness comment claimed "16/16 molresponse ctests pass — the 14 pre-existing suite tests plus the two new pinning tests this rollup adds." The count was real but the composition was wrong: the fixture-gated tests (`test_vbc_spec_equivalence`, `test_tpa_pq_vs_vbc`, `test_tpa_regroup_identities`) were **not among them**. Two stacked reasons, both found this week: (a) our witness builds never set `MADNESS_RESPONSE_FIXTURE_DIR`, so CMake never registered those tests; (b) even with it set, the fixture ground-state archive predated master's August restart-format changes and was rejected as incompatible. The equality gates had not actually executed on this branch until now.

**Fixes.** The fixture archive is regenerated with current code (same geometry, same protocol; stale files preserved in a backup dir), and the full molresponse tier now runs with the fixture configured. Result on the current head: **all tests pass** (only the two madqc scripted tests skip, as expected at this tier), including the three equality gates — `VBC_SPEC_EQUIVALENCE: PASSED`, `PQ_SOURCE_SPEC VALIDATED`, and the M1/M2/M3 regrouping identities.

**New since the last comment** (commits `28ad3b0de`, `46ae3d25f`, `ffcb9f92e`, `1f3eb94da`):
- `test_vbc_pq_terms`: term-resolved comparison of the two quadratic sources (every SourceEntry evaluated separately, aligned slots, per-term norms + contractions). It exposed that the residue source's "extra" family is exactly the B↔C image of its daggered property-Fock family (adjoint identity Ḡ = Gᵀ) — so the symmetrized residue source has exactly the V^BC shape, differing only by the kernel dagger and the pair-density reorientation (both pure HF exchange). Paper §5.6 corrected accordingly.
- `tpa_pq_spec_sym`: the simplified symmetrized builder that exploits the identity (12 two-electron entries instead of 16). The per-ordering builder is kept unchanged as the validation reference; new gate 4 asserts sym == sum-of-orderings on untruncated engine output, tolerance 0.1×thresh (the comparison measures the adjoint identity at MRA operator-apply precision — measured 1.1e-8 at thresh 1e-6). Gate 4 passes.

One process note: the earlier gate-4 draft had its failure flag clobbered by a later `rc` assignment — the same silent-swallowing class this branch's review has been hunting. Fixed (`ffcb9f92e`); the final verdict line now reflects all gates.

robertjharrison:
The concept of seeded state seems to include the idea of solving for an individual interior eigenvalue (i.e., an excited state without first solving for all of the ones below to satisfy orthogonality constraints) which is hard and requires special techniques.   Is this a correct understanding?   If so, how is this interior problem being handled?

ahurta92:
Hi Robert, this PR doesn't address interior states.  The main focus of this PR was restarting "seeding" MADNESS from a Dalton ground/excited-state.  At some point, I was playing around with the idea of starting from a single d-aug-cc-pVQZ response vector, but I don't think it works reliably, and we don't do anything special to make that case work yet.  

One thing the PR actually does is check whether a MADNESS excited-state calculation drifts in frequency and overlap relative to the initial seed-basis-set calculation, warning us when an interior eigenvector might have been missed.

fbischoff (approved):
Those parts outside molresponse look good to me, the other ones are hard to review

PR #786, opened 2026-09-03
Copy target properties into MAD library targets

devreal:
It appears that target_compile_options does not resolve the generator expression but instead copies it verbatim into the target. When the target tries to use the property it fails because the generator expression does not expand correctly. Copying the target expressions explicitly resolves this.

#768 surfaced this issue in TiledArray. Before that, MADworld had no dependency where this would be an issue. Now it MADworld depends on MADmisc, so we should play nice.


robertjharrison:
@devreal ball seems to be in your court.  Trying to close things out so we can get to seeing the release in the distance.

devreal:
@robertjharrison The review comments were addressed already, I just hadn't marked them as resolved.

evaleev:
@devreal I think https://github.com/ValeevGroup/tiledarray/pull/577 should fix the TA problem. Agent suggests there are serious issues with this PR.

devreal:
Thanks for checking and fixing TA. Closing this one.

evaleev:
@devreal I couldn't push to this branch ('Allow edits by maintainers' is off), so I opened #792 instead — it carries your two commits unchanged plus one on top.

Your diagnosis is right and I reproduced the TA failure verbatim. The one thing I'd change: rather than snapshotting the dependency's usage requirements with `get_target_property()`, #792 deletes the copying altogether. The `target_link_libraries(... PUBLIC ...)` immediately below it already propagates everything, so the copy was only ever adding the unevaluable genex to the target's own property. Snapshotting also drops the MPI/json/PaRSEC/LAPACK requirements that reach us through the dependency's *link interface* (`get_target_property` doesn't traverse it), and would miss the LAPACK flags that `src/madness/tensor/CMakeLists.txt` attaches to `MADlinalg` after `add_mad_library(gentensor ...)` runs.

Details and before/after evidence in #792. Happy to fold it back into this PR instead if you'd rather — just flip on maintainer edits.

PR #787, opened 2026-09-04
Probe displacement: only offset where lattice summation demands it, and size the offset from the filter rim

evaleev:
Two follow-ups to #778, stacked on its current head (0348961e4). Both concern `compute_probing_displacement()`; the algorithm's shape is unchanged.

Recall what the probe is for: `do_apply` skips the *entire* boundary-face layer in one shot when `cnorm * op->norm(probe) <= tol/fac`, so the probe has to stand in for the largest-norm displacement in that layer.

### 1. Offset only where lattice summation demands it

The half-cell offset exists to escape lattice equivalence. A face component of `N_d/2` simulation cells folds onto zero only when dimension `d` is **lattice summed** *and* `N_d` is even — when it isn't summed, `maxR == 0`, the image loop in `get_rnlp` never runs, and parity is irrelevant.

`compute_probing_displacement()` branched only on `*box_radius_[d] % 2`, so a free-space range-restricted operator with all-even `N` paid for an offset that bought nothing. Face selection now ranks dimensions by the probe's distance from the center *after* folding — `2^(n-1)` along any summed dimension, `N_d * 2^(n-1)` along an unsummed one — with ties broken toward a dimension needing no offset, then toward the smaller radius. That reproduces the previous choice whenever every dimension is summed, so every expectation in `test_displacements.cc` is preserved.

It matters more than the geometry suggests. The probe sits on the range boundary, where the cutoff makes the kernel fall off far faster than the kernel itself, so a small detour is expensive. Free space, `N=2`, 3D Gaussian, cell [-10,10]³:

| n | PR probe | nearest face point | surface max | PR low by |
|---|---|---|---|---|
| 3 | (8,4,0) `1.11e-18` | (8,0,0) `4.28e-16` | `2.50e-14` | 22,000× |
| 5 | (32,16,0) `4.60e-21` | (32,0,0) `7.22e-18` | `7.22e-18` | 1,571× |
| 7 | (128,64,0) `3.78e-23` | (128,0,0) `6.78e-20` | `6.78e-20` | 1,792× |

Without folding the nearest face point *is* the surface maximum (ratio 1.0 at n=5, n=7) — a tight bound, given up for nothing.

### 2. Size the offset from the filter rim

`2^(n-1)` places the probe at the *farthest*-folded point of the face, the opposite of what it should represent.

The closest surviving surface displacement sits at the inner rim of what the duplicate filter leaves: a displacement is discarded only if it duplicates an already-applied standard displacement, i.e. only if it is within `bmax` along every axis (beyond that it was never in the standard list) **and** no farther than the standard pass reached. So the rim is one box past `min(reach, bmax)` — which does not grow with the level, while the half cell grows as `2^(n-1)`. They coincide only at coarse levels.

`do_apply` now inverts `max_distsq_reached` (already on hand — it is what the validator is built from) into boxes and passes the rim to the range, which clamps it to a half cell, since a step of `2^(n-1)+k` folds back down to `2^(n-1)-k`. A rim of zero means nothing is filtered, so the surface reaches the source box and no offset helps; the conservative face probe is kept and screens nothing, as it should.

Lattice-summed, `N=2`, probe vs. the true maximum over the surviving surface:

| n | reach | rim | half cell | before | after |
|---|---|---|---|---|---|
| 3 | 4 | 4 (clamped) | 4 | conservative | unchanged |
| 5 | 1 | 2 | 16 | **60× low** | **exact** |
| 7 | 1 | 2 | 64 | **198× low** | **4.2× low** |

The residual 4.2× is the probe being a single-axis step while the worst survivor is off-axis — the same order the odd-`N` case has always had (1.03–2.9×), so it is no longer an outlier.

### Also

Minor items from the review that the last batch didn't pick up: `{4,6,4}` was still labelled `{4,6,2}`; two functions shared `test_odds`'s docstring; no trailing newline. And in `compute_probing_displacement()`: the face-dimension assertion sat *after* the `*box_radius_[face_dimension]` deref it guards, the return type carried a pointless top-level `const` while the member wasn't `const`-qualified, the shift in the radius conversion is now parenthesized, and `minimal_second_even`/`minimal_undefined` are renamed to say what they hold.

### Verification

Built `-Werror` under the project's C++20. `test_displacements` 99/99 (38 new checkpoints), `testgconv` exit 0, `testsuite` `passed: true`. Single rank; nothing here touches communication.

https://claude.ai/code/session_012pd8eq8KaHazzd8Le9TFdV


JonathonMisiewicz:
Once again, Claude's comments are utterly unreadable. I'll decipher its gibberish and make the appropriate commits to the mother PR. "the cutoff makes the kernel fall off far faster than the kernel itself," seriously?

PR #788, opened 2026-09-05
BoxSurfaceDisplacementRange: choose the probe offset in real space

evaleev:
Follow-up to #778, targeting its branch so only the fix shows.

**Per-face probes (third commit).** A single probing displacement gated the whole surface: `for_each` evaluated its norm once and, if negligible, skipped every face. That is sound only if the probe is at least as close as the nearest kept surface point on *any* face, which no face ranking can guarantee once faces sit at different real-space distances (mixed-parity lattice-summed radii, lattice summation along some axes only, anisotropic cells). Now every face has its own probe (`probing_displacement(d)` / `probing_displacements()`), `do_apply` probes each face, skips the negligible ones via `skip_face(d)`, and skips the surface only if every face screens. A skipped face is simply not iterated: a box is visited iff it lies on at least one non-skipped face, independent of face order. The face ranking is gone.

**Real-space offset (second commit).** The offset of an even lattice-summed face was computed in boxes from the widest cell axis while the offset axis was chosen by box radius, so in anisotropic cells the probe could sit farther than the nearest kept surface point. `probe_offset_radius` is replaced by `StandardDisplacementsReach` (max real distance squared plus the cell widths that produced it, carried by the validator since the fourth commit); the range inverts it per axis and picks the offset axis with the least real-space offset. The widths are the same ones `default_real_distance_squared` uses, so probe and validator agree. Omitting the reach means nothing is known to be filtered, so the on-site probe is used.

*No reproducer for this one.* The defect is in the contract between `funcimpl.h` and the range, not inside either: at the base the range takes a single box count, `funcimpl.h` derives it from the widest cell axis, and the range applies it along the finite axis with the smallest radius. Neither side is wrong on its own terms, and cell widths only enter the interface with this fix, so no test against the base API can fail. The anisotropic cases in `test_standard_reach` pin the corrected behavior. A failing demonstration would need an end-to-end `apply` of a range-restricted lattice-summed operator in an anisotropic periodic cell, compared against an unscreened reference (there is no switch to disable the probe), with tolerances tuned so the over-screening error stands out from truncation error; I have not attempted that.

**Reproducers first.** The first commit (063dbe60c) adds only two tests to `test_displacements.cc` and fails against the unchanged iteration code (identical to master): CI run https://github.com/m-a-d-n-e-s-s/madness/actions/runs/34061462957. The fixes follow in the next two commits and turn it green.

**Two pre-existing iteration bugs** surfaced by the new tests and fixed in the third commit:
- Without a validator the first surface point was never yielded (`advance_till_valid` advanced unconditionally in that branch).
- For lattice-summed radii of 2 or more half cells the box is wider than the cell, so the range of box indices between its two faces, [-r+t+1, r-t-1] in the integer box coordinates the iterator works in, spans more than one period; after a face was processed the later faces iterated over that whole range and thus re-enumerated boxes that are lattice-equivalent to the layers just processed (the 3×3 edge boxes of a 2D N={2,2} surface twice, 189 duplicates for 3D N={2,2,2} at level 3). Bounds of a processed lattice-summed dimension now shrink to the complement of its layer classes within one period; hollowness is defined modulo the period too.

**Review follow-ups (fourth commit).** The iteration bounds are lattice-aware from initialization (one period ending at the top layer along a lattice-summed dimension), which removes the equivalence-class clipping in `reset_along_dim` and makes the post-face shrink and the hollowness test uniform. `BoxSurfaceDisplacementValidator` now owns `StandardDisplacementsReach` (a namespace-scope struct) and computes its distances from it; the range takes an optional validator object instead of a callable plus a separate reach, reads the reach from it, and checks that the lattice-summation flags agree. The validator no longer takes the (unused) kernel range or a distance callable. Probe docs reworded: a finite-radius dimension has a pair of faces, served by one probe, which errs toward the largest norm.

**Re-review and audit fixes (fifth and sixth commits).** The probe sits on the innermost layer r-t and, for a lattice-summed face, is folded into the cell like the yielded displacements. Faces whose layers are all filtered out (entirely outside a finite domain) are skipped instead of tripping an assert. The validator clips `bmax` to 2^n-1 on every axis whenever any axis is lattice summed, matching `make_disp_periodic`. A 64-bit shift at level 0 is guarded. Tests cover mixed-axis lattice summation, faces outside the domain, and the validator's `bmax`. Two pre-existing issues found by the audit are recorded in a PR comment and left for follow-up: the standard displacement list is ordered with unit cell widths, and the `Keep` policy on an unrestricted axis uses the absolute cell as the face window.

**Review nits.** Unused `<limits>` dropped, `<cmath>` added, `compute_probing_displacement` is `const` and returns by value, Doxygen ordering and a typo fixed, cell-width check is an always-on `MADNESS_CHECK_THROW`.

**Tests.** `test_displacements`: per-face probe expectations, anisotropic-cell cases, the half-cell cap, a face-skipping test against the "on at least one remaining face" rule, and uniqueness of the surface modulo the lattice for lattice-summed cases. `test_displacements`, `testgconv`, and `testpercoul` pass locally (Release, PaRSEC, macOS).

https://claude.ai/code/session_01RX8v3398TLeTzqAghZ5ivQ

### DIFF
--- a/src/madness/mra/displacements.h
+++ b/src/madness/mra/displacements.h
@@ -39,9 +39,9 @@
 
 #include <algorithm>
 #include <array>
+#include <cmath>
 #include <functional>
 #include <iterator>
-#include <limits>
 #include <optional>
 #include <tuple>
 #include <utility>
@@ -340,6 +340,13 @@ namespace madness {
       /// the validator should normally be a BoxSurfaceDisplacementFilter object. anything else is probably a hack.
       using Validator = std::function<bool(Level, const PointPattern&, std::optional<Displacement>&)>;
 
+      /// Real-space extent of the standard (short-range) displacements that BoxSurfaceDisplacementValidator
+      /// filters out of the surface; used to place the probing displacement just outside of it.
+      struct StandardDisplacementsReach {
+        double max_distsq;                    ///< max real distance squared reached by the standard displacements (see Key::real_distsq_bc)
+        std::array<double, NDIM> cell_width;  ///< real-space width of the simulation cell along each axis, as used to compute `max_distsq`
+      };
+
     private:
       using BoxRadius = std::array<std::optional<Translation>, NDIM>;  // null radius = unlimited size
       using SurfaceThickness = std::array<std::optional<Translation>, NDIM>;  // null thickness for dimensions with null radius
@@ -355,8 +362,7 @@ namespace madness {
       Hollowness hollowness_;    ///< does box contain non-surface points along each dimension?
       Periodicity is_lattice_summed_;  ///< which dimensions are lattice summed?
       Validator validator_;      ///< optional validator function
-      std::optional<Translation> probe_offset_radius_; ///< least N such that being N boxes away from origin guarantees
-                                                       ///   the point was not a short-range point already considered
+      std::optional<StandardDisplacementsReach> standard_reach_;  ///< what the validator filters out; null if nothing is known to be filtered
       Displacement probing_displacement_;  ///< displacement to a nearby point on the surface; it may not be able to pass the filter, but is sufficiently representative of the surface displacements to allow screening with isotropic kernels
 
       /**
@@ -686,19 +692,21 @@ namespace madness {
        * @param surface_thickness Surface thickness in each dimension, measured in number of addl. boxes *on each half* of the surface box proper. Omit for dim `i` if and only if omitted in `box_radius`
        * @param is_lattice_summed whether each dimension is lattice summed; along lattice summed dimensions only one side of the box is iterated over.
        * @param validator Optional filter function (if returns false, displacement is dropped; default: no filter); it may update the displacement to make it valid as needed (e.g. map displacement to the simulation cell)
+       * @param standard_reach the real-space extent of the standard/short-range displacements that `validator` discards
+       *        from the surface (see BoxSurfaceDisplacementValidator); it is used to place the probing displacement
+       *        just outside of that region. Omit if nothing is known to be filtered out: the surface then reaches all
+       *        the way in to `center`, and the probe falls back to the on-site displacement, which screens nothing.
        * @pre `surface_radius[d]>0 && surface_thickness[d]<=surface_radius[d]`
-       * @param probe_offset_radius the smallest displacement magnitude, in boxes, that `validator` does *not* discard as a duplicate of the standard/short-range displacement list, Pass 0 to signal that nothing is filtered out (the surface then reaches all the way in to `center` and no probe can screen it). Omit if unknown, in which case the half-simulation-cell offset is used.
-       *
        */
       explicit BoxSurfaceDisplacementRange(const Key<NDIM>& center,
                                            const std::array<std::optional<std::int64_t>, NDIM>& box_radius,
                                            const std::array<std::optional<std::int64_t>, NDIM>& surface_thickness,
                                            const array_of_bools<NDIM>& is_lattice_summed,
                                            Validator validator = {},
-                                           std::optional<Translation> probe_offset_radius = {})
+                                           std::optional<StandardDisplacementsReach> standard_reach = {})
           : center_(center), box_radius_(box_radius),
             surface_thickness_(surface_thickness), is_lattice_summed_(is_lattice_summed), validator_(std::move(validator)),
-            probe_offset_radius_(probe_offset_radius) {
+            standard_reach_(std::move(standard_reach)) {
         // initialize bounds
         bool has_finite_dimensions = false;
         const auto n = center_.level();
@@ -774,7 +782,7 @@ namespace madness {
       }
 
     private:
-      const Displacement compute_probing_displacement() {
+      Displacement compute_probing_displacement() const {
         // Large boxes we must consider are both those near the center (because 1/r is large
         // for small r), and near the box radius (because going from 1/r to 0 is a sharp change).
         // The probe displacement is a way to screen out cases where the box radius is negligible.
@@ -806,8 +814,8 @@ namespace madness {
         //    even though they're actually 0, to account for the offset we'll need to add.
         // 2. Is an offset needed? Avoiding it is preferred.
         // 3. Choose the smallest N possible.
-        // Requirement (3) would be better formulated in real-space, but the expected
-        // bound improvement doesn't justify expanding the argument signature.
+        // Requirement (3) would be better formulated in real-space (cell widths are available via
+        // standard_reach_, when given), as is done for the offset below.
         const auto sort_key = [&](size_t d) {
           const auto N = *box_radius_[d];
           return std::make_tuple(is_lattice_summed_[d] ? Translation(1) : N, face_origin_is_center(d), N);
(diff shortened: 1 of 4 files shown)


src/madness/mra/displacements.h
```diff
@@ -835,43 +843,56 @@ namespace madness {
...
+        // Cap the offset at half a cell. If our dimension is lattice-summed, it's even, and half a cell
+        // is where it's furthest from the origin. Else, half a cell is the furthest away we can
+        // guarantee we can displace to, in the case of an open dimension and the center_ is the origin.
+        const Translation half_cell = Translation(1) << (n-1);
+        const Translation bmax = Displacements<NDIM>::bmax_default();
+        const auto offset_along = [&](size_t d) -> Translation {
+          const double width = standard_reach_->cell_width[d];
+          MADNESS_ASSERT(width > 0);
+          const Translation reach = 1 + static_cast<Translation>(std::sqrt(standard_reach_->max_distsq) / width);
+          return std::min(std::min(reach, bmax) + 1, half_cell);
```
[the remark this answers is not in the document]
evaleev: Agreed; it is now an always-on `MADNESS_CHECK_THROW(width > 0, ...)` (cb46c1578).

src/madness/mra/funcimpl.h
```diff
@@ -5199,29 +5199,26 @@ template<size_t NDIM>
...
-              const std::size_t d0 = (op->particle() == 1) ? 0 : NDIM - opdim;
-              double widest = 0.;
-              for (std::size_t d = d0; d != d0 + opdim; ++d) widest = std::max(widest, cell_width(d));
-              const Translation reach = 1 + static_cast<Translation>(std::sqrt(*max_distsq_reached) / widest);
-              probe_offset_radius = std::min<Translation>(reach, Displacements<opdim>::bmax_default()) + 1;
-            } else  // no validator => no treatment of short-range point. let surface handle everything
-              probe_offset_radius = 0;
+              std::array<double, opdim> widths;
+              for (std::size_t d = 0; d != opdim; ++d) widths[d] = cell_width(d);
+              standard_reach = typename SurfaceRange::StandardDisplacementsReach{*max_distsq_reached, widths};
```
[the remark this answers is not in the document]
evaleev: This is deliberate. The widths must be the ones the validator uses to compute `max_distsq_reached`, i.e. `default_real_distance_squared`, which calls `Key<opdim>::real_distsq_bc` with the full `FunctionDefaults<NDIM>::get_cell_width()` and therefore reads widths `0..opdim-1` for either particle. Using the back widths here would make the probe and the validator disagree for particle 2, which is exactly what this PR fixes. If the particle-2 distance function is ever changed to use the back widths, this line must follow it (see the N.B. in the code); the 6D periodic case is not exercised by any test today.

src/madness/mra/displacements.h
```diff
@@ -355,9 +362,9 @@ namespace madness {
       Hollowness hollowness_;    ///< does box contain non-surface points along each dimension?
       Periodicity is_lattice_summed_;  ///< which dimensions are lattice summed?
       Validator validator_;      ///< optional validator function
-      std::optional<Translation> probe_offset_radius_; ///< least N such that being N boxes away from origin guarantees
-                                                       ///   the point was not a short-range point already considered
-      Displacement probing_displacement_;  ///< displacement to a nearby point on the surface; it may not be able to pass the filter, but is sufficiently representative of the surface displacements to allow screening with isotropic kernels
+      std::optional<StandardDisplacementsReach> standard_reach_;  ///< what the validator filters out; null if nothing is known to be filtered
+      std::array<std::optional<Displacement>, NDIM> probing_displacements_;  ///< for each finite-radius dimension, a displacement to a nearby point on the face normal to it; it may not be able to pass the filter, but is sufficiently representative of that face's displacements to allow screening with decaying kernels
```
JonathonMisiewicz: This comment is technically incorrect.

1. A finite-radius dimension has _two_ faces normal to it.
2. We also want the displacement to err on the side of having a larger norm. That's the whole motivation for the last couple of Claude PRs. See my larger comment.

src/madness/mra/displacements.h
```diff
@@ -492,23 +484,53 @@ namespace madness {
           }
         }
 
-        /// Perform advance, repeating if you are at a filtered point
+        /// Excludes the layers of the face normal to `dim` from the faces that remain to be processed,
```
JonathonMisiewicz:
```suggestion
        /// Excludes the layers of the faces normal to `dim` from the faces that remain to be processed,
```

evaleev:
@JonathonMisiewicz reproducers for both iteration bugs are now the first commit of the branch (063dbe60c), adding only two tests to `test_displacements.cc`:

- surface uniqueness *modulo the lattice* for lattice-summed cases (9 duplicates for 2D N={2,2}, 189 for 3D N={2,2,2} at level 3), and
- iteration with an accept-all validator vs. no validator (191 vs 192 displacements for 2D N={1,1}).

The iteration code at that commit is unchanged from master, so the failure is master's. CI on that commit alone: https://github.com/m-a-d-n-e-s-s/madness/actions/runs/34061462957 (expected red in the Threads cells that run the unit tests). The two fix commits follow on top and the branch tip is green locally (`test_displacements`, `testgconv`, `testpercoul`).

src/madness/mra/displacements.h
```diff
@@ -355,9 +362,9 @@ namespace madness {
       Hollowness hollowness_;    ///< does box contain non-surface points along each dimension?
       Periodicity is_lattice_summed_;  ///< which dimensions are lattice summed?
       Validator validator_;      ///< optional validator function
-      std::optional<Translation> probe_offset_radius_; ///< least N such that being N boxes away from origin guarantees
-                                                       ///   the point was not a short-range point already considered
-      Displacement probing_displacement_;  ///< displacement to a nearby point on the surface; it may not be able to pass the filter, but is sufficiently representative of the surface displacements to allow screening with isotropic kernels
+      std::optional<StandardDisplacementsReach> standard_reach_;  ///< what the validator filters out; null if nothing is known to be filtered
```
JonathonMisiewicz: Note that for this technology to work, the validator now *has* to represent a radius. We've been implicitly assuming this before, but let's explicitly document this. It's worth considering whether we should construct the validator from `standard_reach_`.

src/madness/mra/displacements.h
JonathonMisiewicz: It may say "the face," but you can have two faces. They just lattice sum onto each other, should we be using lattice sums. This is important for the anisotropy that Florian works with.

JonathonMisiewicz:
Some historical context on this saga before I get to my notes on the actual code (which looks very good):

In the early days of experimenting with the surface, we concluded that the value of the probe norm was the same across the surface. A much better zero-order approximation is that the value is inversely proportional to r. (I recall observing that norm did decay as distance increased, but I never fit that to a curve.)

After #732 (which was mindful of this and tried to be greedy with the probe) became incredibly complicated, #778 initially proposed a new, much simpler algorithm that ignored all of those fine distinctions and used the poorer-quality approximation that surface effects are constant along the surface. Numerical testing indicated that although there was a detectable accuracy loss with this code, it was below epsilon, so it was good enough.

During the edits on #778, Claude's changes introduced some changes to optimize back to choosing the surface point with the largest norm. Bizarrely, Claude's comments showed zero mention of this, which is why I had to rewrite them all from scratch.

Now, I could tell the changes were an improvement but not complete. Since even my unoptimized code was within epsilon, I let it slide without pressing the point. However, it seems that with the benefit of my comments, Claude was able to figure out the problem. This PR tries to optimize the probe point, which takes us back to the complexity that I was hoping to avoid in the first place... but at least #732 introduced a simpler starting point for the problem, so these changes are actually readable.

src/madness/mra/displacements.h
JonathonMisiewicz: Again, shouldn't this be faces?

src/madness/mra/displacements.h
JonathonMisiewicz: We also want the displacement to err on the side of having a larger norm among the (broadly) representative displacements. That's the whole motivation for the last couple of Claude PRs. See my larger comment.

JonathonMisiewicz (commented):
The code overall looks good, but I wonder if refactoring `unprocessed_bounds` could simplify the complexity added by fixing the lattice-sum bug. With the displacement code, the simpler, the better...

In the old code, we didn't mind periodicity at all when defining `unprocessed_bounds` but set it equal to the un-lattice-summed-box and trusted other parts of the code to resolve duplicates and convert to equivalence classes. It's no coincidence that the newly added code to fix the duplicate issue so closely resembles the conversion from dishis even further.placements to equivalence_classes in `reset_along_dim`...

It's worth checking if we can make our definition of `unprocessed_bounds` itself lattice-summation-aware. If so, we could likely eliminate our special handling of lattice summed dimensions elsewhere, which would simplify this code.

src/madness/mra/displacements.h
evaleev: Done in 828d78ee8, the other way around: the validator now owns the reach. `BoxSurfaceDisplacementValidator` takes `StandardDisplacementsReach` (now a namespace-scope struct), computes the distances from it with the same `Key::real_distsq_bc` that orders the standard displacements, and exposes it; the range takes an optional validator object instead of a callable plus a separate reach, reads the reach from it, and checks that the lattice-summation flags agree. The free-form `Validator` callable is gone, so the assumption is now enforced by the types rather than documented.

src/madness/mra/displacements.h
evaleev: Reworded in 828d78ee8: "the faces normal to d, the pair of hyperplanes at -radius and +radius (which lattice summation folds onto each other)". One probe still serves both since the real-space distance depends only on the displacement magnitude along each axis.

src/madness/mra/displacements.h
evaleev: Reworded to "faces" here and in the other places that said "the face" (828d78ee8).

src/madness/mra/displacements.h
evaleev: Agreed, and that is what the per-face probe does (nearest kept point on that face, with the offset minimized in real space). The member and accessor docs now say so explicitly: "among the displacements those faces contribute it errs toward the largest norm" (828d78ee8).

evaleev:
@JonathonMisiewicz re the `unprocessed_bounds` refactor: done in 828d78ee8, and it simplifies things as you predicted. The bounds are lattice-aware from initialization: along a lattice-summed dimension they are one period ending at the top layer, so each equivalence class appears exactly once. With that, the equivalence-class clipping in `reset_along_dim` is gone (only the layers of a lattice-summed face still need clipping to one period, in case they are thicker than the cell), `exclude_face` uniformly drops the layer boxes from the ends of the bounds (both ends, or the top end only if lattice summed), and hollowness is uniformly "more boxes in the bounds than in the layers". Net effect on `displacements.h` is about even in line count because the validator moved above the range, but the lattice-summed special cases went from three to one.

The same commit makes the validator own the reach (your other comment) and rewords "face" to "faces". `test_displacements`, `testgconv` and `testpercoul` pass locally.

evaleev:
re: complexity: yes, this is more complex, but for high aspect ratios I'd rather be safe than simple.

src/madness/mra/funcimpl.h
```diff
@@ -5102,14 +5102,23 @@ template<size_t NDIM>
...
-                  op->norm(key.level(), probing_displacement, source);
-              if (cnorm * opnorm <= tol / fac) {
+              bool any_face_survives = false;
+              const auto &probing_displacements = displacements.probing_displacements();
+              for (std::size_t d = 0; d != opdim; ++d) {
+                if (!probing_displacements[d]) continue;  // no face normal to an unlimited dimension
+                const double opnorm =
+                    op->norm(key.level(), *probing_displacements[d], source);
+                if (cnorm * opnorm <= tol / fac)
+                  displacements.skip_face(d);
```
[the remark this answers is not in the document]
evaleev: Correct, and fixed in fbfe31021: the probe now sits on the innermost layer, r-t. A follow-up (4991f910a) also folds the probe of a lattice-summed face into the cell, to the representative nearest to the source, as the validator already does for the displacements it yields; without that the operator norm, which only sums over a few lattice images, saw a probe up to N/2 cells away.

evaleev:
Follow-up to Copilot's re-review plus an independent audit of the range, validator and the `do_apply` call site (fbfe31021, 4991f910a):

**Fixed**
- Probe on the innermost layer r-t rather than r (Copilot).
- Probe of a lattice-summed face folded into the cell, as the yielded displacements are; `op->norm` only sums a few lattice images, so a probe N/2 cells away was underestimated for N ≥ 3.
- A face all of whose layers are filtered out (every layer outside a finite domain, which a range of one cell or more on a non-periodic function produces) tripped a `MADNESS_ASSERT` in `reset_along_dim`, and in release builds walked the filtered face point by point. Such faces are now skipped like finished ones. This predates the PR.
- The validator clipped `bmax` to 2^n-1 only along lattice-summed axes, but `make_disp_periodic` clips it along every axis whenever any axis is summed; at small levels a displacement with 2^n-1 < |l| ≤ bmax along an unsummed axis was taken for a standard one and dropped unprocessed. Predates the PR.
- A 64-bit shift at level 0 in the validator (unreachable in production).

**Tests added**: mixed-axis lattice summation in the uniqueness and face-skipping tests (Copilot's suppressed ask), faces and whole surfaces outside the domain, the validator's `bmax` at a small level and at level 0, and the probe expectations moved to the innermost folded layer.

**Not done: the end-to-end regression Copilot asked for** (range-restricted lattice-summed operator in an anisotropic cell driving `do_apply`). There is no switch to disable probe screening, so the reference would have to come from a tighter threshold or a hand-assembled surface sum, with tolerances tuned to separate over-screening from truncation error. The PR description records this.

**Found, pre-existing, out of scope for this PR** (both worth an issue):
- `Displacements::widths` defaults to 1 and `set_width` is only called from `testopdir.cc`, so the standard displacement list is ordered with unit widths while `do_apply` measures shells and `max_distsq_reached` with the real cell widths. For anisotropic cells the shells fragment, the early break can leave displacements within the reach unprocessed, and the validator then discards their surface images as duplicates. Isotropic cells are unaffected.
- For an unrestricted axis under the `Keep` policy (function periodic along it, operator not lattice-summed along it) the face window is the absolute cell [0, 2^n), so the box adjacent across the wrap is never enumerated while its far alias is.

PR #792, opened 2026-09-07
cmake: stop stamping dependency usage requirements into MAD* target properties

evaleev:
Supersedes #786. Squashed to a single commit, since #786's two commits are entirely undone by the change below; @devreal is credited as co-author, and the diagnosis is his.

### The bug #786 found is real

`add_mad_library()` copied each MAD dependency's `INTERFACE_COMPILE_DEFINITIONS` / `INTERFACE_INCLUDE_DIRECTORIES` / `INTERFACE_COMPILE_OPTIONS` into the dependent target as `$<TARGET_PROPERTY:...>` generator expressions. Those are stored *verbatim* in the dependent's own property, so tooling that reads the property raw cannot evaluate them. TiledArray's `DetectMADNESSConfig.cmake` feeds `MADworld`'s `INTERFACE_INCLUDE_DIRECTORIES` into a standalone `try_compile` that knows nothing about `MADmisc`, and since #768 made `MADworld` depend on `MADmisc` that configure dies:

```
CMake Error at .../CMakeFiles/CMakeScratch/TryCompile-2LiXYf/CMakeLists.txt:11 (include_directories):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:MADmisc,INTERFACE_INCLUDE_DIRECTORIES>

  Target "MADmisc" not found.
```

I reproduced this by running TA's `detect_MADNESS_configuration()` macro verbatim against a `add_subdirectory()`'d MADNESS tree.

### Why this PR fixes it differently

#786 keeps the copying but snapshots the *resolved* values with `get_target_property()`. That removes the genex, but it is lossy and order-dependent:

- **Lossy.** `$<TARGET_PROPERTY:tgt,INTERFACE_INCLUDE_DIRECTORIES>` resolves transitively over `tgt`'s link interface; `get_target_property()` returns only the target's own property. `MADworld`'s own include dirs contain no MPI/json/PaRSEC paths — those arrive via `MPI::MPI_CXX`, `nlohmann_json::nlohmann_json` and `parsec` in its `INTERFACE_LINK_LIBRARIES`. Snapshotting drops them from `MADtensor`/`MADlinalg`/`MADmra`/`MADchem`'s own properties, which is the same class of standalone-property failure the PR set out to fix.
- **Order-dependent.** The snapshot captures whatever the dependency carries at macro-call time. `src/madness/tensor/CMakeLists.txt` calls `add_mad_library(gentensor ... "linalg" ...)` *before* it attaches `LAPACK_INCLUDE_DIRS` / `LAPACK_COMPILE_OPTIONS` / `LAPACK_COMPILE_DEFINITIONS` to `MADlinalg`, so `MADgentensor` would silently miss them under `-DENABLE_GENTENSOR=ON`.

The copying was never needed in the first place. Immediately after it, the macro already does `target_link_libraries(${targetname} PUBLIC ${deptargetname})`, and that PUBLIC link gives both the target's consumers and `MAD${_name}-obj` (which forwards `$<TARGET_PROPERTY:MAD${_name},INCLUDE_DIRECTORIES>` & co. at the top of the macro) the dependency's full, transitively-resolved usage requirements. So the copy contributed nothing to the build graph — its only observable effect was to stamp an unevaluable genex into the target's own property.

**This PR deletes the copying and relies on the link interface.** Net effect versus master: 19 lines removed from `AddMADLibrary.cmake`.

It also fixes two latent target-name bugs uncovered by removing the copies:

- the header-only branch of `add_mad_library()` linked the bare component name (`-lmisc`) rather than `MAD<dep>`;
- `add_mad_hdr_library()` guarded on `if(TARGET ${_dep})`, never true since every component target is `MAD<dep>` — the entire body was dead code (which is also why the pre-existing `PUBLIC` scope on an `INTERFACE` library, a hard CMake error, was never hit).

### Verification

macOS/arm64, AppleClang 17, Ninja, Release, C++20, PaRSEC backend, OpenMPI 5.0.9.

- **TA repro, before/after.** TA's `DetectMADNESSConfig.cmake` run verbatim against this tree: fails on `master` with the error above; configures clean with this PR.
- **Nothing is lost.** Dumped every `MAD*` target's *resolved* `INTERFACE_{INCLUDE_DIRECTORIES,COMPILE_DEFINITIONS,COMPILE_OPTIONS,LINK_LIBRARIES}` and every `MAD*-obj`'s resolved `INCLUDE_DIRECTORIES`/`COMPILE_DEFINITIONS` via `file(GENERATE)`, for `master` versus this branch. Identical on every target; only ordering shifts. MPI, nlohmann_json, PaRSEC and Threads all still arrive everywhere they did before.
- **Exported targets are clean.** `madness-targets.cmake` now contains zero `$<TARGET_PROPERTY:MAD*>` expressions, with `MADmisc` carried in `INTERFACE_LINK_LIBRARIES` as it should be.
- **Full build** of libraries, apps and examples: 623/623, exit 0.
- **`check-short-madness`**: 233 tests, 231 pass. The two failures are unrelated to this change — `madness/test/mra/testdiff1Db/run` aborts identically on a pre-existing pre-#786 build tree, and `znemo/test_energy.py` merely timed out under `MAD_NUM_THREADS=2`, passing in 41 s at 8 threads.

Not covered locally: `-DENABLE_GENTENSOR=ON` and `-DENABLE_MPI=OFF`.

https://claude.ai/code/session_01Q4FnRms4DMd5M9pVo1fvvX

### DIFF
--- a/cmake/modules/AddMADLibrary.cmake
+++ b/cmake/modules/AddMADLibrary.cmake
@@ -66,14 +66,24 @@ macro(add_mad_library _name _source_files _header_files _dep_mad_comp _include_d
       endif(${_dep}_is_mad_hdr_lib)
 
       if(TARGET ${deptargetname})
-        target_compile_definitions(${targetname} PUBLIC
-            $<TARGET_PROPERTY:${deptargetname},INTERFACE_COMPILE_DEFINITIONS>)
-        target_include_directories(${targetname} PUBLIC
-            $<TARGET_PROPERTY:${deptargetname},INTERFACE_INCLUDE_DIRECTORIES>)
-        target_compile_options(${targetname} PUBLIC
-            $<TARGET_PROPERTY:${deptargetname},INTERFACE_COMPILE_OPTIONS>)
+        # Propagate the dependency's usage requirements through the link interface only.
+        #
+        # Do NOT also copy the dependency's INTERFACE_{COMPILE_DEFINITIONS,INCLUDE_DIRECTORIES,
+        # COMPILE_OPTIONS} into ${targetname}'s own properties. The link below already gives
+        # both ${targetname}'s consumers and MAD${_name}-obj (which forwards
+        # $<TARGET_PROPERTY:${targetname},INCLUDE_DIRECTORIES> & co. above) the dependency's
+        # full, transitively-resolved usage requirements, so the copy adds nothing to the
+        # build graph -- but it does leave a $<TARGET_PROPERTY:${deptargetname},...> genex
+        # sitting in ${targetname}'s own property. Downstream tooling that reads those
+        # properties raw (e.g. TiledArray's DetectMADNESSConfig.cmake, which feeds
+        # MADworld's INTERFACE_INCLUDE_DIRECTORIES into a standalone try_compile that knows
+        # nothing about MADmisc) then dies with 'Target "MADmisc" not found.'
+        # Copying the *resolved* values instead is no better: get_target_property() does not
+        # traverse the dependency's own link interface, so the MPI/json/PaRSEC/LAPACK
+        # requirements that reach us through it would be silently dropped, and the snapshot
+        # would capture only what the dependency happens to carry at macro-call time.
         if (${_dep}_is_mad_hdr_lib)
-          target_link_libraries(${targetname} INTERFACE ${_dep})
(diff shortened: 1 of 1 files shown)


devreal (approved):
Thanks for the fix @evaleev!

PR #796, opened 2026-09-09
PCM

fbischoff:
…LE_PCM off

ENABLE_PCM previously required a preinstalled PCMSolver that only the hand-rolled FindPCM module could locate -- in practice, one whose prefix the user passed in PCM_ROOT_DIR. Give it the same three-step treatment external/dftd3.cmake gives simple-dftd3, with a source build as the last resort so asking for PCM is enough to get it:

1. find_package(PCMSolver CONFIG), which picks up the PCMSolver::pcm target that PCMSolver >= 1.2 installs. Preferred over the module because the config carries the transitive link deps a static libpcm needs (zlib, the Fortran runtime); FindPCM finds a header and a library and leaves those to luck.
2. FindPCM, unchanged, for older or hand-laid-out installations.
3. cmake/modules/FindOrFetchPCMSolver.cmake, which fetches the pinned tag and builds it as a subproject.

All three normalize onto PCM_FOUND / PCM_LIBRARIES / PCM_INCLUDE_DIRS, so src/madness/CMakeLists.txt and src/madness/chem/CMakeLists.txt are untouched. ENABLE_PCM now defaults OFF, so none of this runs unless asked for; MADNESS_FETCH_PCMSOLVER (default ON) turns off just step 3.

Two things the source build needs that are not obvious:

- enable_language(Fortran). PCMSolver's cavity generator (PEDRA) is Fortran, and CMake does not carry a language's implicit link libraries across an OBJECT library -- which is how PCMSolver assembles pcm-static -- so a Fortran-free C++ consumer links short of __gfortran_*. Enabling the language lets CMake see Fortran in pcm-static's link closure and add the runtime itself. Done inside FindOrFetchPCMSolver, not at the top of the project: external/lapack.cmake runs earlier and deliberately probes Fortran symbols from C, and no configure that skips the fetch pays for it. The runtime is additionally pinned onto pcm-static's *install* interface, since an installed-tree consumer acts on IMPORTED_LINK_INTERFACE_LANGUAGES only if it has Fortran enabled, which a pure C++ consumer of MADNESS has no reason to do. Scoping that to INSTALL_INTERFACE keeps it off in-tree link lines, where it would only earn a duplicate-libraries warning on every executable.

- Header staging. PCMSolver arranges its headers under PCMSolver/ only as part of install(); its build-tree include interface points at the flat source dirs. chem/pcm.h asks for <PCMSolver/pcmsolver.h>, so the three headers that entails are staged into the build tree, and pcm-static's INTERFACE_INCLUDE_DIRECTORIES is repointed at that -- which also keeps PCMSolver's own src/ and include/ trees off the header search path of every MADNESS target that links PCM.

Prerequisites (Fortran compiler, Boost headers, zlib) are probed before fetching; a missing one prints which and leaves PCM off rather than failing configure, since a deck that asks for pcm aborts with an explanatory message anyway.

Upstream v1.3.0 is from 2020 and does not build on a current toolchain. cmake/patches/pcmsolver-v1.3.0.cmake carries four hunks, each idempotent and each a hard error if its context is gone, so moving MADNESS_TRACKED_PCMSOLVER_TAG fails loudly instead of silently building something untested: cmake_minimum_required 3.3 -> 3.5 (CMake >= 4 refuses below 3.5); C++11 -> C++14 (Boost >= 1.87 odeint uses generic lambdas, and SphericalDiffuse.cpp pulls odeint in); the bundled Eigen 3.3.2 Transpositions.h trt.derived() bug (breaks clang >= 11, fixed upstream in 3.3.5); and a rename of their `update_version` target, too generic a name for a tree MADNESS may itself be a subproject of. PCMSolver's own generate-config-hpp race is fixed from our side with an explicit add_dependencies.

madness-config.cmake.in re-finds PCMSolver for consumers on the two paths that need it -- an installed config, or our own build, whose targets are exported from the build tree (PCMSolver exports only its install tree) and whose config lands under the MADNESS prefix. The FindPCM path is already baked into the exported target as an absolute path and needs nothing.

The feature-summary entry is named "PCM solvation", not "PCM": that name is taken by the FindPCM package, and a FeatureSummary entry that is both a package and a feature is reported as neither -- it went missing from the summary in every state, enabled included.

Verified on macOS/arm64 (AppleClang 17, gfortran 15, CMake 4.3.1) and on Linux/x86-64 (GCC 13.3, Boost 1.92, CMake 4.3.1): fetch, patch, libpcm.a, chem/pcm.cc, MADchem, and an executable link with a clean link line (no duplicate-libraries warning) on both. On Linux, moldft --geometry=water --dft="xc=hf; pcm_data=water" builds a GePol cavity (3 spheres, Bondi-Mantina, 296 tesserae) and runs IEFPCM to -76.07669392 vs -76.06792844 in gas phase, a -5.5 kcal/mol solvation shift. The skip path was exercised for real on a host without Boost headers (configure clean, PCM off), the CONFIG path against a conda PCMSolver, and ENABLE_PCM off/on/off toggling without a re-fetch. Reconfigures are idempotent, and both export(EXPORT madness) and install(EXPORT madness) resolve PCMSolver::pcm, with the Fortran runtime present in the installed PCMSolverTargets-static.cmake and absent from the build-tree one.


fbischoff:
@ahurta92  I’ve revived the pcm solvation model, it works for the ground state. At some point I had it running also for excited states, so you could implement that functionality too, but if not you would probably want to throw an error if both response and pcm are requested at the same time to prevent silent wrong results.

ahurta92:
Awesome, @fbischoff, I'll take a look at it today.  Do you have any input I can use to test it?

fbischoff:
sure, just add to the dft block "pcm_data = YourSolvent" , with YourSolvent e.g. "water" or "acetone" or another common solvent

fbischoff:
@robertjharrison @ahurta92 pcm is now ready for review, I've checked against psi4 that also uses pcmsolver

fbischoff:
@robertjharrison @hborchert @ahurta92  will force-merge the PCM PR tomorrow unless there are objections

PR #801, opened 2026-09-11
Displacements: keep the ordering of the standard displacements in sync with the cell

### LINKED ISSUE #799: Displacements: standard displacement list is ordered with unit cell widths, not the FunctionDefaults cell
Found while auditing #788; predates it.

`Displacements<NDIM>::widths` (displacements.h:71) is a static tensor that defaults to 1 and is only changed by `Displacements<NDIM>::set_width`, whose sole callers are two lines in `testopdir.cc`. Both displacement lists (`disp`, `disp_periodic[n]`) are sorted by that tensor through `cmp_keys` / `cmp_keys_periodic`.

`FunctionImpl::do_apply` measures shells with the real cell widths, `real_distsq_bc(op->lattice_summed(), FunctionDefaults<NDIM>::get_cell_width())` (funcimpl.h ~5073), and the shell loop assumes the list is monotone in that metric: it breaks as soon as a whole shell contributes nothing, and hands the distance of the last processed shell to `BoxSurfaceDisplacementValidator` as the reach of the standard displacements.

For a cubic cell the two metrics agree up to a constant. For an anisotropic cell they do not, e.g. widths (1, 10, 10): the list puts (0,2,0) before (3,0,0), but in real space (0,2,0) is at distance 10 and (3,0,0) at 2. Consequences:

1. The standard loop can break early: a shell at real distance 10 that contributes nothing stops the loop while nearer displacements sit later in the list, unprocessed. This is a plain loss of contributions, independent of range restriction.
2. For range-restricted kernels, the validator then drops surface displacements within `max_distsq_reached` as duplicates on the premise that everything within that reach was processed, which after an early break is false.

Fix: keep `Displacements::widths` in sync with the `FunctionDefaults` cell (re-sort when the cell changes), and add a test asserting that `get_disp` is sorted by `real_distsq_bc` with the current cell widths after an anisotropic `set_cell`.

evaleev:
Closes #799.

**Problem.** `Displacements<NDIM>::widths` defaulted to 1 and `set_width` was only called from `testopdir.cc`, so the standard displacement lists were ordered with unit cell widths, while `FunctionImpl::do_apply` measures shells with the real `FunctionDefaults` cell widths. For an anisotropic cell the list is not monotone in the metric the shell loop assumes (e.g. widths (1,10,10): (0,2,0) precedes (3,0,0) in the list but is five times farther in real space). The early break can then leave nearer displacements unprocessed, and for range-restricted kernels the surface validator drops their images as duplicates. Cubic cells are unaffected.

**Reproducer first.** The first commit only adds a test that the lists are ordered by real-space distance in the current cell, for the cubic default and for an anisotropic cell set after the lists were built. It fails against master on the anisotropic checks (CI on that commit alone, eed663de6: https://github.com/m-a-d-n-e-s-s/madness/actions/runs/34609551533, expected red in the Threads cells).

**Fix.** `FunctionDefaults::recompute_cell_info` now calls `Displacements::set_width`, so the lists follow the cell whenever it is set, at setup time and off the task pool. The two headers depend on each other (`Displacements` reads the boundary conditions from `FunctionDefaults`), so both forward-declare the other's class, `displacements.h` no longer includes `funcdefaults.h`, and `funcdefaults.h` includes `displacements.h` at its end, after `FunctionDefaults` is complete. `set_width` copies the widths instead of aliasing the cell tensor, and its doc now says it must not run while operators are being applied.

**Review follow-ups (third commit).** `displacements.h` includes `funcdefaults.h` again so it stays self-contained (the include guards make the header cycle safe in either order); `indexit.h`, which it includes first, now includes `madness/world/vector.h`, which it was missing. `set_width` is a no-op when the widths are unchanged, so re-setting the same cell at startup does not reorder the lists; `set_cell` / `set_cubic_cell` document that they must be called with no operator application in flight, since the lists are reordered in place.

**Not covered.** For a 6D operator acting on one particle, `do_apply` measures with the first three widths of the 6D cell while `Displacements<3>` is ordered by the 3D cell; the two agree when the 6D cell is the replicated 3D cell, which is the only case in use.

**Verification** (Release, PaRSEC, macOS): `test_displacements` (323 checkpoints), `testgconv`, and `testopdir` pass at the tip; the test-only commit fails as intended.

Rebased onto master after #802 brought #788 in; the ordering test now sits alongside the #788 tests.

### DIFF
--- a/src/madness/mra/displacements.h
+++ b/src/madness/mra/displacements.h
@@ -34,7 +34,8 @@
 #define MADNESS_MRA_DISPLACEMENTS_H__INCLUDED
 
 #include <madness/mra/indexit.h>
-#include <madness/mra/funcdefaults.h>
+#include <madness/mra/key.h>
+#include <madness/misc/array_of_bools.h>
 #include <madness/tensor/tensor.h>
 
 #include <algorithm>
@@ -48,6 +49,8 @@
 #include <vector>
 
 namespace madness {
+
+    template <std::size_t NDIM> class FunctionDefaults;  // funcdefaults.h includes this header at its end
 
     // How should we treat destinations "extra" to the [0, 2^n) standard domain?
     enum class ExtraDomainPolicy {
@@ -287,8 +290,11 @@ namespace madness {
           MADNESS_PRAGMA_CLANG(diagnostic pop)
         }
 
+        /// Sets the cell widths used to order the displacements by real-space distance, and reorders them.
+        /// Called by FunctionDefaults::recompute_cell_info whenever the cell changes.
+        /// @warning not thread safe: must not be called while operators are being applied
         static void set_width(const Tensor<double>& width) {
-          widths = width;
+          widths = copy(width);
           if (!disp.empty()) {
             std::sort(disp.begin(), disp.end(), cmp_keys);
           }
--- a/src/madness/mra/displacements.h
+++ b/src/madness/mra/displacements.h
@@ -34,8 +34,7 @@
 #define MADNESS_MRA_DISPLACEMENTS_H__INCLUDED
 
 #include <madness/mra/indexit.h>
-#include <madness/mra/key.h>
-#include <madness/misc/array_of_bools.h>
+#include <madness/mra/funcdefaults.h>
 #include <madness/tensor/tensor.h>
 
 #include <algorithm>
@@ -49,8 +48,6 @@
 #include <vector>
 
 namespace madness {
-
-    template <std::size_t NDIM> class FunctionDefaults;  // funcdefaults.h includes this header at its end
 
     // How should we treat destinations "extra" to the [0, 2^n) standard domain?
     enum class ExtraDomainPolicy {
@@ -290,10 +287,15 @@ namespace madness {
           MADNESS_PRAGMA_CLANG(diagnostic pop)
         }
 
-        /// Sets the cell widths used to order the displacements by real-space distance, and reorders them.
-        /// Called by FunctionDefaults::recompute_cell_info whenever the cell changes.
-        /// @warning not thread safe: must not be called while operators are being applied
+        /// Sets the cell widths used to order the displacements by real-space distance, and reorders them
+        /// if the widths changed. Called by FunctionDefaults::recompute_cell_info whenever the cell is set.
+        /// @warning reorders the lists in place: must not be called while operators are being applied
+        ///          (see FunctionDefaults::set_cell)
         static void set_width(const Tensor<double>& width) {
+          MADNESS_ASSERT(width.ndim() == 1 && width.size() == NDIM);
+          bool changed = widths.ndim() != 1 || widths.size() != NDIM;
+          for (std::size_t i = 0; !changed && i != NDIM; ++i) changed = widths(i) != width(i);
+          if (!changed) return;
           widths = copy(width);
           if (!disp.empty()) {
             std::sort(disp.begin(), disp.end(), cmp_keys);
--- a/src/madness/mra/funcdefaults.h
+++ b/src/madness/mra/funcdefaults.h
@@ -48,6 +48,7 @@
 #include <optional>
 
 namespace madness {
+    template <std::size_t NDIM> class Displacements;  // defined in displacements.h, included at the end of this file
     template <typename T, std::size_t NDIM> class FunctionImpl;
 
     /// The maximum wavelet order presently supported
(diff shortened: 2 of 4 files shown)


src/madness/mra/displacements.h
```diff
@@ -34,7 +34,8 @@
 #define MADNESS_MRA_DISPLACEMENTS_H__INCLUDED
 
 #include <madness/mra/indexit.h>
-#include <madness/mra/funcdefaults.h>
+#include <madness/mra/key.h>
+#include <madness/misc/array_of_bools.h>
```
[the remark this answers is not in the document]
evaleev: Agreed. displacements.h includes funcdefaults.h again (cdbe5a363); funcdefaults.h keeps the end-of-header include, and the include guards make the cycle safe in either order. Checking this with syntax-only compiles of each header alone also showed indexit.h was not self-contained (it uses `Vector` without including it, and displacements.h includes it first); fixed in the same commit. All three headers now compile on their own.

src/madness/mra/funcdefaults.h
```diff
@@ -149,6 +150,9 @@ namespace madness {
             cell_min_width = cell_width.min();
             rcell_width = copy(cell_width);
             for (std::size_t i=0; i<NDIM; ++i) rcell_width(i) = 1.0/rcell_width(i);
+            // the standard displacements used to apply operators are ordered by real-space distance,
+            // which depends on the cell; keep them in sync (see Displacements::set_width)
+            Displacements<NDIM>::set_width(cell_width);
```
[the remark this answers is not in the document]
evaleev: Partly addressed, partly documented (cdbe5a363). `set_width` is now a no-op when the widths are unchanged, so re-setting the same cell, which is what `set_defaults` and most applications do at startup, does not touch the lists. A genuine cell change while `do_apply` tasks are in flight cannot be fenced from `FunctionDefaults`, which is process-global static state without a `World`; and such a change already renders existing functions and operators invalid (the setters say so, and CLAUDE.md requires the cell to be set before functions are constructed). The setters now state the quiescence requirement explicitly.

src/madness/mra/displacements.h
```diff
@@ -287,8 +287,16 @@ namespace madness {
...
         }
 
+        /// Sets the cell widths used to order the displacements by real-space distance, and reorders them
+        /// if the widths changed. Called by FunctionDefaults::recompute_cell_info whenever the cell is set.
+        /// @warning reorders the lists in place: must not be called while operators are being applied
+        ///          (see FunctionDefaults::set_cell)
         static void set_width(const Tensor<double>& width) {
-          widths = width;
+          MADNESS_ASSERT(width.ndim() == 1 && width.size() == NDIM);
+          bool changed = widths.ndim() != 1 || widths.size() != NDIM;
```
JonathonMisiewicz, line 296: When could `widths.ndim() == 1` possibly fail? I understand how `widths.size() != NDIM` could be true or false - we may have changed the dimension of the space we're considering. But `widths.ndim() == 1` should be an invariant. It is one dimension of size ndims.

src/madness/mra/displacements.h
```diff
@@ -287,8 +287,16 @@ namespace madness {
...
+          for (std::size_t i = 0; !changed && i != NDIM; ++i) changed = widths(i) != width(i);
```
JonathonMisiewicz, line 297: Do we need to make this a fuzzy check, or is it safe to assume that no change means equality? My intuition is that this is _probably_ safe, but that _probably_ is not good enough for what we're doing.

JonathonMisiewicz:
I could _swear_ that was called in the actual code. It's unclear to me how this broke, but thank you for fixing it regardless. Address two minor comments, then LGTM.

src/madness/mra/displacements.h
evaleev: It can't; `widths` is only ever assigned `Tensor<double>{NDIM}`, the scalar fill `widths = 1` (shape-preserving), or `copy(width)` after asserting `width` is 1D of size NDIM. Made it an assertion of the invariant instead of a condition (1c7320d3a).

src/madness/mra/displacements.h
evaleev: Exact is the safe direction here. The comparison only decides whether to skip the reorder, so the two failure modes are asymmetric: an unnecessary reorder (widths differ by 1e-16, nothing changes) costs a sort at setup and is harmless; a missed reorder is exactly the bug this PR fixes, and a fuzzy tolerance could in principle miss a change that flips the order of two displacements at a near-tie. Left exact and said so in a comment (1c7320d3a).

PR #804, opened 2026-09-14
Dirac: More Parameters to Restore Old Behavior

EricaCMitchell:
## Description

`DFdriver` had several behaviors that were fixed in the code and could not be changed from the input deck: the sparsity threshold used by `mul_sparse()`, the quantity that ends the SCF loop, and the treatment of a closed-shell guess. It also equated "closed shell" with "Kramers restricted", which doubled the density of a closed-shell unrestricted run, and it wrote restart archives with no format version, so an old archive was read as if it were current.

This branch adds the missing input keywords (`thresh_mul`, `convergence_criteria`), separates `closed_shell` from `Krestricted`, versions the restart archive and rejects unversioned files with an actionable message, and makes the end of the SCF loop say why it stopped. The convergence logic moves into a header-only helper (`DFConvergence.h`) so it can be tested, and the `dirac` application gets its first unit-test target (`test_dirac.cc`, label `short`).

**The new keywords do not move the results.** With the defaults, the branch reproduces the energies that PR #789 reports. With `thresh_mul 0.0` and `convergence_criteria bsh_residual`, it **reproduces the energies of master as it was before PR #789.**

### Input knobs

- [X] [f79847327](https://github.com/m-a-d-n-e-s-s/madness/commit/f79847327b51b4e24baa174a805d8189058abf8b) `dirac:` adds the `thresh_mul` keyword to `DFParameters`, threaded through to the `mul_sparse()` calls in `DF.cc`. The old behavior is recovered by setting it to the refinement threshold.
- [X] [aaed34421](https://github.com/m-a-d-n-e-s-s/madness/commit/aaed3442180b29f20441ab6cd4531196a1695845) `dirac:` adds `convergence_criteria`. `bsh_residual` (default) keeps the old test on the maximum BSH residual; `energy_density_residual` also requires the relative total-energy change `<= thresh` and the density change `<= dconv * nelec`, with the BSH limit relaxed to `100 * thresh`. Both tests run after the orbital update, where the energy and density of the iteration exist. An unknown value is rejected at parse time.
- [X] [675a1e2f9](https://github.com/m-a-d-n-e-s-s/madness/commit/675a1e2f9cbcdd7b3126974438394f3d3aa75cf6) `dirac:` fixes keyword-order dependence in `DFParameters::read()`. `thresh` overwrote `dconv` and `thresh_mul` as it was parsed, so an explicit value survived only if it came after `thresh`. `read()` now tracks whether each was set and applies the `thresh` default after the block ends.

### Kramers restriction

- [X] [198d9492d](https://github.com/m-a-d-n-e-s-s/madness/commit/198d9492d9e01b41efda19c183e81d0b183faec9) `dirac:` stops treating `closed_shell` as an alias for `Krestricted`. The density factor in `DF::diagonalize`, `DF::iterate`, and `DF::solve_occupied` is now `2` only under `Krestricted`, and `InitParameters` expands a spin-restricted `moldft` guess into both members of each Kramers doublet when `Krestricted` is off.

### Convergence helper and tests

- [X] [39ba5c1d2](https://github.com/m-a-d-n-e-s-s/madness/commit/39ba5c1d29f226f836d86a800290d45bf947a48e) `dirac:` moves the convergence test into `DFConvergence.h` (`DFConvergenceCriterion`, `DFConvergenceMetrics`, `df_iteration_converged`) and fixes the combined criterion, which compared the current energy against itself instead of against the previous iteration. Adds `test_dirac.cc` and registers it with `add_unittests(dirac ... "unittests;short")`.
- [X] [c93aa4f2c](https://github.com/m-a-d-n-e-s-s/madness/commit/c93aa4f2c0754c0c630e823131b2e7a29fea8be1) `dirac:` adds boundary cases to the helper tests: the exact BSH tolerance, and the fallback to an absolute energy change when the current total energy is zero (the relative form would divide by zero).

### Restart archives

- [X] [599a63d72](https://github.com/m-a-d-n-e-s-s/madness/commit/599a63d724f7addb8edbfe78c33b2ef6f33e7ac1) `dirac:` adds `DFRestart.h`, which writes `DF_RESTART_VERSION` as the archive's first datum and validates it on read. An unversioned archive starts with a `double`, so the type cookie mismatch identifies it; `read_df_restart_version` catches that and returns 0, and `require_supported_df_restart` throws with a message naming the recovery path. Rank 0 reads the version from `input.local_archive()` and broadcasts it, so every rank throws together instead of waiting inside `wrap_load`'s broadcast.

### Termination reporting

- [X] [98f5efe0e](https://github.com/m-a-d-n-e-s-s/madness/commit/98f5efe0e82d980f697afd131596d500438ce44b), [7b447e8a4](https://github.com/m-a-d-n-e-s-s/madness/commit/7b447e8a40414473c2a759144b86d7646855d9d3) `dirac:` adds `DFStopReason` / `df_stop_message` and prints one of them when the loop ends. Hitting `max_iter` previously ended `DF::solve_occupied` silently, so an unconverged run looked like a converged one; it now prints `WARNING: maximum iterations reached without convergence` and the iteration count. `README.md` gains a "Termination" section saying that such a run is not converged, and that the restart archive is still written.

### Comments

- [X] [a71f40245](https://github.com/m-a-d-n-e-s-s/madness/commit/a71f402459f802d7baf109be6e2a3d918749e52c) `dirac:` rewrites the comments this branch adds to follow ASD-STE100 Simplified Technical English. Removes passive voice, present perfect, the banned modals `would` and `could`, and a semicolon. A comment in `DFConvergence.h` now states why `df_convergence_criterion_from_string` returns a bool: `MadnessException` stores the message as a bare `const char*`, so a message built from the keyword dangles after the throw.

## AI

Claude Code wrote the unit tests in `test_dirac.cc`, found the combined-criterion energy bug and the keyword-order dependence in `DFParameters::read()`, and drafted the `README.md` parameter table entries and the "Termination" section.

## Reviewer notes (evaleev)

Two user-visible consequences that should go into the release notes:

- **Cost change for existing decks.** Without `Krestricted`, a closed-shell run from a spin-restricted `moldft` guess now iterates 2N spinors instead of N. Per-iteration cost roughly doubles and the exchange build scales worse. Add `Krestricted` to keep the old cost.
- **Checkpoint-format break.** DF restart archives written before this PR are rejected with no migration path. Regenerate them from a `moldft` archive.


evaleev (commented):
Read the full diff plus the surrounding code in `DF.cc`, `InitParameters.h`, and the archive layer. No correctness problems found: the `fac = Krestricted ? 2 : 1` change is consistent through `diagonalize`, both density rebuilds in `iterate`, the kinetic/Coulomb/exchange sums, and `solve_occupied`; the existing `MADNESS_CHECK(spinrestricted == Krestricted)` on restart keeps a Krestricted archive from being read into an unrestricted run with the new factor; the version probe relies on the archive cookie check throwing `MadnessException`, which it does; and CI's Release+Threads cells built and ran `test_dirac` under `check-short-madness`.

Two things worth stating explicitly in the PR description / merge commit, since users will hit them:

1. **Cost change for existing decks.** A deck without `Krestricted` that starts from a spin-restricted moldft guess now iterates 2N spinors instead of N. Per-iteration cost roughly doubles and the exchange build scales worse than that. The README covers it, but a one-line hint in the log when a closed-shell guess runs Kramers-unrestricted ("add `Krestricted` to halve the cost") would save people a surprise.

2. **Checkpoint-format break.** Every DF restart archive written before this PR becomes unreadable, with no migration. That is intentional and documented, but it belongs in the release notes.

Copilot's two suppressed nits are both valid; I've left them inline.

src/apps/dirac/DFRestart.h
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/// Throws unless \c version is a DF restart format this build can read.
+inline void require_supported_df_restart(const unsigned int version) {
+    MADNESS_CHECK_THROW(version != 0,
+        "unversioned DF restart archive; start from a moldft archive or "
+        "rerun with the older DFdriver");
+    MADNESS_CHECK_THROW(version == DF_RESTART_VERSION,
+        "unsupported DF restart archive version; start from a moldft archive "
+        "or regenerate it with this DFdriver");
```
evaleev: This branch fires when the archive is *newer* than this build, so "regenerate it with this DFdriver" points at the DFdriver that just rejected it. Suggest: "DF restart archive was written by a newer DFdriver; use that DFdriver, or start from a moldft archive".

src/apps/dirac/README.md
```diff
@@ -157,6 +182,13 @@ end
 ```
 Ensure the previous run did not specify `no_save`.
 
+DF restart archives begin with an explicit format version field. Archives written before this
+field was introduced carry no version and are rejected with an actionable error; they are not
+migrated. Regenerate them by restarting from a `moldft` archive. The incompatibility is detected
+in both directions: an older `DFdriver` pointed at a new archive reads the version field where it
```
evaleev, line 188: True on one rank only. With several ranks, the older DFdriver's `ParallelInputArchive` throws on rank 0 inside `wrap_load`, and the other ranks wait in its broadcast, so the run hangs rather than aborting. Suggest softening to: rank 0 reports an archive type mismatch; multi-rank runs may hang at that point.

src/apps/dirac/DF.cc
```diff
@@ -1738,12 +1754,12 @@ void DF::solve_occupied(World & world)
...
-          if(DFparams.Krestricted || closed_shell){
-               if(closed_shell) print("\nSolving for ", Init_params.num_occupied, " doubly-occupied orbitals\n------------------------------\n");
-               else print("\nSolving for ", Init_params.num_occupied-1, " doubly-occupied, 1 singly-occupied orbitals\n------------------------------\n");
+          if(DFparams.Krestricted){
+               if(closed_shell) print("\nSolving for ", Init_params.num_occupied, " Kramers-restricted orbitals\n------------------------------\n");
+               else print("\nSolving for ", Init_params.num_occupied-1, " Kramers-restricted open-shell orbitals\n------------------------------\n");
           }
           else{
-               print("\nSolving for ", Init_params.num_occupied, " single-occupied orbitals\n------------------------------\n");
+               print("\nSolving for ", Init_params.num_occupied, " Kramers unrestricted orbitals\n------------------------------\n");
```
evaleev, line 1762: Since a closed-shell guess now lands here by default, consider adding a line such as `(closed shell: add Krestricted to the input to halve the cost)` when `closed_shell` is true. Cheap, and it explains the 2N orbital count to anyone comparing against an older run.

src/apps/dirac/DFRestart.h
EricaCMitchell: Fixed

src/apps/dirac/DF.cc
EricaCMitchell: Fixed

src/apps/dirac/README.md
EricaCMitchell: Fixed

src/apps/dirac/InitParameters.h
```diff
@@ -106,6 +118,7 @@ namespace madness{
 
                     //For now assume spin-restricted means closed shell in moldft
                     closed_shell = spinrestricted;
+                    if(world.rank()==0 && !Krestricted) print("closed shell: add Krestricted to the input to halve the cost");
```
evaleev: This prints for a spin-unrestricted moldft guess too, where `closed_shell` is false and the advice does not apply. Gate it on the flag just set:
```suggestion
                    if(world.rank()==0 && closed_shell && !Krestricted) print("closed shell: add Krestricted to the input to halve the cost");
```

src/apps/dirac/InitParameters.h
```diff
@@ -353,6 +367,7 @@ namespace madness{
                     numbeta = numalpha;
                }
                closed_shell = !have_beta;
+               if(world.rank()==0 && !Krestricted) print("closed shell: add Krestricted to the input to halve the cost");
```
evaleev: Same here: `have_beta` (open shell) still reaches this line, so the hint is wrong for that case.
```suggestion
               if(world.rank()==0 && closed_shell && !Krestricted) print("closed shell: add Krestricted to the input to halve the cost");
```

src/apps/dirac/DFRestart.h
```diff
@@ -0,0 +1,69 @@
...
+///
+/// Earlier archives carry no version field, and this code rejects them. The read
+/// goes through the type cookie that the BinaryFstream archives write.
+///
+/// Current Version:
+/// 0) DF restart format version (unsigned int)
+/// 1) Total energy (double)
+/// 2) Krestricted (boolean)
+/// 3) closed_shell (boolean)
+/// 3) number of occupied orbitals (int)
```
evaleev: Nit: item numbering repeats `3)`, and this field is written as `unsigned int` (`Init_params.num_occupied`), not `int`.

src/apps/dirac/DFRestart.h
```diff
@@ -0,0 +1,69 @@
...
+/// 4) orbital energies (vector of doubles)
+/// 5) box size (double)
+/// 6) wavelet order (int)
+/// 7) molecule (molecule)
+/// 8) occupied orbitals as complex functions
+///
+/// /note v1 introduced format version
```
evaleev: Nit: Doxygen wants `\note`, not `/note`.

evaleev (commented):
Re-reviewed the three follow-up commits (bb0d58b30, 3b745fbf3, 558e380a9). The `DFRestart.h` message and the README wording are fixed as suggested, the existing `UnsupportedVersionsExplainRecovery` test still matches the new strings, and CI passed on 558e380a9 including the dirac unit test.

One small defect in the new log hint (inline): both `print` calls fire on `!Krestricted` alone, so an open-shell guess also prints "closed shell: add Krestricted". Plus two doc nits in the new layout comment. Once the hint is gated on `closed_shell` I'm happy to approve.

Note: Copilot's re-review on this head failed twice (an internal error, then the requester's quota), so there is no second automated pass on these commits.

evaleev (approved):

PR #809, opened 2026-09-15
BoxSurfaceDisplacementRange: center the window of an unrestricted axis on the source under the Keep policy

### LINKED ISSUE #800: BoxSurfaceDisplacementRange: Keep policy on an unrestricted axis uses the absolute cell as the face window
Found while auditing #788; predates it.

For an axis with no kernel range the surface window is the absolute cell [0, 2^n) (displacements.h, `box_[d] = {0, 2^n-1}` for unrestricted dimensions), i.e. displacements [-C, 2^n-1-C] for a source at C. That is right under the Discard policy (destinations outside the cell are invalid anyway) and under Translate (any full period of equivalence classes is the same). It is wrong under Keep, i.e. the function is periodic along that axis but the operator is not lattice-summed along it: destinations outside the cell are legitimate and `neighbor()` wraps them, but the kernel is not periodic, so displacements l and l ± 2^n are different contributions to the same wrapped box.

Example: 2D, range restricted along x only, function periodic in y, operator not summed in y, level 4, source at y = 0. The window enumerates y-displacements 0..15. The box one step across the wrap, displacement -1, carries a kernel value comparable to displacement 0 and is never enumerated; its alias at displacement +15 is enumerated instead, landing on the same destination with the kernel evaluated fifteen boxes out. The near contribution is lost.

Fix: under Keep, make the window one period of displacements centered on the source, [-2^{n-1}, 2^{n-1}-1] (what the standard list does with its symmetric bmax). The range now holds the validator (#788), which knows the per-axis policy, so this is a small change. Test: 2D {*,2} range with a Keep validator and a source at y = 0 must enumerate a displacement with y = -1.

Trigger is narrow (range restriction without lattice summation along a periodic axis), hence low priority.

evaleev:
Fixes #800.

**Terminology.** Keep, Translate and Discard are the values of the `ExtraDomainPolicy` enum in `src/madness/mra/displacements.h` (introduced in #649). `BoxSurfaceDisplacementValidator` assigns one per axis from its two constructor flags, `is_infinite_domain` and `is_lattice_summed` (filled in `FunctionImpl::do_apply` from the operator's `func_domain_is_periodic()` and `lattice_summed()`), and it says what to do with a destination box outside the simulation cell `[0, 2^n)`:

- **Discard**: the axis is neither periodic nor lattice summed. Such destinations are invalid and are filtered out. Most computations.
- **Keep**: the function is periodic along the axis but the operator is *not* lattice summed along it. The destination is valid; the caller wraps it into the cell, the displacement is left as is (the kernel is not periodic, so `l` and `l ± 2^n` are different contributions).
- **Translate**: the operator is lattice summed along the axis. Displacements that differ by a period are equivalent, so the displacement itself is shifted to bring source and destination into the cell.

Along an axis with no kernel range the surface spans one period of destinations, which was always the absolute cell `[0, 2^n)`. That is right under Discard (destinations outside the cell are invalid) and under Translate (any period holds the same equivalence classes), but wrong under Keep, i.e. when the function is periodic along the axis and the operator is not lattice summed along it: destinations outside the cell are legitimate and wrapped by the caller, but the kernel is not periodic, so displacements `l` and `l ± 2^n` are different contributions to the same wrapped box. For a source near the cell boundary the box one step across the wrap was never enumerated; its alias a period away was enumerated instead, with the kernel evaluated a period out, and the near contribution was lost.

## Change

- Under Keep the window of an unrestricted axis is the period centered on the source, `[-2^{n-1}, 2^{n-1}-1]`, as for the standard displacements. Discard and Translate keep the cell, as does a range without a validator (nothing is known then).
- `BoxSurfaceDisplacementValidator` exposes its per-axis policy (`domain_policy(d)`), which the range consults.
- The probe offset along an unrestricted axis is capped at a half cell, so it still fits the window; only its comment changed.

## Test

`test_displacements`: a 2D range restricted along x only with a source at `y = 0` (level 4) must enumerate `y = -1` and not its alias `y = 2^n-1` under Keep, and exactly the cell under Discard. The three Keep checks fail on master and pass with the fix; the Discard check passes on both, guarding against centering unconditionally.

Verified in a RelWithDebInfo build with `ASSERTION_TYPE=assert` and `MADNESS_WERROR=ON`: all of `test_displacements` passes, and `testgconv` (whose RP2 case is the Keep path in 1D) passes as a regression check.


JonathonMisiewicz:
Claude has failed to convince me there's a bug here. Why is it imperative that the standard window contain all the largest displacements? The way I remember thsi code working, displacements are generated on center and then wrapped to the correct cell in the window. The window is not used to generate the displacements themselves. Unless there have been changes to the displacement logic that I'm not aware of, this is entirely unnecessary.

evaleev:
first off, this is for unused path (periodic function, non-lattice-summed kernel), only useful for testing. Here's what Claude had to say in response:

The finite-radius axes are generated on center (box_ = center ± r), and the caller does wrap the destination
(FunctionImpl::neighbor). The unrestricted axis is different: box_[d] = {0, 2^n-1} is absolute, the iterator
walks point[d] from 0 to 2^n-1, and the displacement is point - center. Nothing remaps it along a Keep axis
(the validator only remaps under Translate). The wrap applies to dest, but op->norm and op->apply receive the
enumerated displacement. So for a source at y = 0 the wrapped box at y = 15 gets the kernel at +15, and the
kernel at -1 is never applied. The new test run on master shows y-displacements exactly 0..15 under Keep for
that source. Under Translate the two are the same value, which is why this only bites without lattice summation.

JonathonMisiewicz:
Claude still isn't convincing me. I'll run some tests myself.

JonathonMisiewicz:
Propose that we close. See my response on issue #800.

PR #810, opened 2026-09-16
WorldContainer: Make Replication Linear and Idempotent

EricaCMitchell:
## Description

`WorldContainer::replicate()` sent each entry once per rank turn, so replication traffic grew with the rank count instead of staying at one copy per entry. This PR makes rank replication transmit every original entry exactly once, makes a repeated call send no container entries when every rank's replica remains valid (it still performs one `gop.min` reduction to agree on that condition, plus the requested trailing fence), and documents when node replication is a safe policy.

### Background

`do_replicate` walked the container with `begin()`/`end()` while it inserted into that same container inside the loop. Rank `k` therefore broadcast not only its own shard, but every shard it had already received from ranks `0..k-1`. With `N` global nodes over `P` ranks, the total sent was `N * (P + 1) / 2` entries rather than `N` — an excess factor that grows without bound in `P`.

Two further costs multiplied it. The old code broadcast each entry as a separate key and a separate value. `broadcast_serializable` (`worldgop.h:774`) issues two blocking tree broadcasts per call, so replication required `4 * entries + 2 * P` latency-bound tree traversals.

Measured on master, four 3-D functions at `k=8`, `thresh=1e-6`, `N=1828` nodes, Open MPI on one machine at `MAD_NUM_THREADS=1`:

| Ranks | Entries sent | Excess factor | Replicate time |
|---|---|---|---|
| 1 | 0 | — | 0.0001 s |
| 2 | 2784 | 1.52 | 0.302 s |
| 4 | 4768 | 2.61 | 0.971 s |
| 8 | 8736 | 4.78 | 1.867 s |

The measured entry counts match the `N * (P + 1) / 2` model, which predicts 2742, 4570 and 8226. At 27 μs per broadcast, the traversal count accounts for most of the 2-rank time — the operation is latency-bound, not bandwidth-bound.

### Transient memory: peak grows to about 2× the largest rank shard

This PR increases the peak transient memory of `do_replicate`. The old code sent one entry at a time, so it held one serialized entry at a time. The new code sends one root's whole shard in one `broadcast_serializable` call:

- **On the root rank:** the `entries` vector, plus a `BUFLEN` byte buffer (`worldgop.h:786`) that holds the serialized shard. For `Tensor` values the vector is a shallow copy, so the buffer is the main cost: about 1× the shard.
- **On each receiving rank:** the same `BUFLEN` buffer, plus the deserialized `vector<pair<key, value>>` with new, deep tensor data. Both exist at the same time until `broadcast_serializable` deletes the buffer. The peak is about 2× the root's shard, in addition to the memory that the rank already uses.

The memory is released before the next root's broadcast starts, so the peak depends on the largest single rank shard, not on the total container size. Chunked broadcasts (`worldgop.cc:218`) prevent MPI `int` overflow, but they do not decrease this peak.

For 3-D production functions the shard is small and the increase is not significant. For 6-D work (MP2/CC2, `-DENABLE_GENTENSOR=ON`), one rank's shard can be many GB. A run that fitted in memory before this PR can then fail with out-of-memory during `replicate()`.

### Application-level effect: MPQC MRA-MCSCF

Measured in MPQC (`173ae7292e`), Be atom, Dirac/X2C, `L=30`, eps=1e-5, on one MacBook Air
(M3). Each cell is the median of 3 runs, with the two tags built separately and the runs
shuffled and interleaved in one session. Two steps were timed: the purified atomic orbitals
built once per run, and the reconstruction of the atomic spinor partners done once per MCSCF
iteration. Both steps call `replicate()`.

| Metric | Setup | `f2d0a38d` | this branch | speed-up [95% CI] |
|---|---|---|---|---|
| Purified atomic orbitals | 1 rank × 4 threads | 5.19 s | 5.20 s | 1.00× [0.96, 1.02] |
| Spinor partners, per call | 1 rank × 4 threads | 3.10 s | 3.11 s | 1.00× [0.98, 1.04] |
| Purified atomic orbitals | 2 ranks × 3 threads | 8.30 s | **5.21 s** | **1.59× [1.57, 1.65]** |
| Spinor partners, per call | 2 ranks × 3 threads | 3.92 s | **2.82 s** | **1.39× [1.36, 1.40]** |
| Total wall time | 2 ranks × 3 threads | 393.5 s | 369.4 s | 1.07× [0.99, 1.08] |

With 2 ranks the branch makes the purified orbitals 1.59× faster and the spinor-partner
reconstruction 1.39× faster than `f2d0a38d`. With 1 rank, where `do_replicate` returns at once,
the branch costs nothing: both steps land within 1.5% of `f2d0a38d` and the confidence
intervals include 1. The branch also removes the penalty for adding a second rank — under
`f2d0a38d` reconstruction gets 26% slower going from 1 rank to 2 (3.10 → 3.92 s), under this
branch it gets 9% faster (3.11 → 2.82 s). Energies are identical between the two commits
(−14.6197110503 Eh at eps=1e-5), so the branch changes speed only.

The confidence intervals are bootstraps over 3 runs per cell, on one machine and one molecule.
Total wall time is dominated by other work and is given for context only.

## Details

- [X] [728224379](https://github.com/m-a-d-n-e-s-s/madness/commit/728224379ff210be7ec58d119e63fef21ab85848) `world:` `do_replicate` (`worlddc.h`) records the local keys of each rank before it communicates. It then broadcasts one packed `std::vector<std::pair<keyT, valueT>>` per root, so each entry crosses the wire one time. The broadcast count falls from `2N(P+1) + 2P` to `2P`, and a one-rank world returns at once. Peak transient memory grows to about 2× the largest rank shard; see the section above. The new `test_replication.cc` counts the output serializations and the packed payloads, and `CMakeLists.txt` registers it with a 2-rank MPI variant.
- [X] [642db0453](https://github.com/m-a-d-n-e-s-s/madness/commit/642db04532d423ba6ee5979641b50d9d07b68c36) `world:` `WorldContainerImpl` gets a `rank_replication_complete_` flag. `replicate()` reduces the flag with `world.gop.min`, then does the requested fence and returns if all ranks agree. Thus no rank skips a broadcast that another rank enters. `clear()`, `reset_pmap_to_local()`, `replicate_on_hosts()`, and both redistribute phases clear the flag. New tests cover a full clear, a clear on one rank, a map-only pmap reset, and redistribution before and after replication. The docs tell the caller that queued tasks must not change the container during the call.
- [X] [498eaede4](https://github.com/m-a-d-n-e-s-s/madness/commit/498eaede482d267f36ca21769efd343cc2aee960) `mra:` `Function::replicate(type, fence)`, `replicate(fence)`, and `replicate_on_hosts(fence)` (`mra.h`) now give the ownership rule for each policy. Under node replication the lowest rank on each host owns all coefficient nodes, other ranks route container access such as `find` to that rank, and `eval_local_only` returns false on them because it does no communication. `test_memory_measurement.cc` asserts that `eval_local_only` succeeds only on the host-primary rank. No call site changes its policy.

## Verification

- `madness/test/{world/test_replication, mra/test_memory_measurement}` and their `_mpi2` variants: 4/4 pass.
- `MAD_NUM_THREADS=1 mpiexec -np 3 --bind-to none ./build/src/madness/world/test_replication` and `-np 4`: exit code 0.
- MADWorld short suite (`ctest -L short -R madness/test/world`): 24/24 pass.
- Additionally, in a separate build directory `-DENABLE_MPI=OFF -DMADNESS_BUILD_MADWORLD_ONLY=ON`: configures, builds, and passes the single-process test (`madness/test/world/test_replication/run`). The change adds no raw `MPI_*` calls — everything goes through `World::gop`.
- clang-tidy: no diagnostics on any changed line.
- `git diff --check f2d0a38de HEAD`: clean.
- **Not tested: packed replication on the node-replication path.** `replicate_on_hosts` calls `do_replicate(world_primary)`. On one host, `world_primary` has one rank, and `do_replicate` returns at once, so the packing code does not run on that path. Every run above used one host. A 2-host run of `test_memory_measurement_mpi2`, with one rank per host, runs that code.

### Notes for reviewers

- Replication remains a collective distribution conversion. A repeated call does not reconcile replicas that other code modified independently; the new documentation states this. Per-key edits to a replicated container must be made the same way on every rank; `clear()` resets replication, so the next `replicate()` call redistributes from scratch.
- Node replication stays opt-in and benefits from the corrected shared helper.
- The PR makes no change to MRA checkpoint serialization or `FunctionDefaults`.

## AI
- Claude Code diagnosed the `N * (P + 1) / 2` scaling defect, implemented the fix and the idempotency tracking, wrote `test_replication.cc`, and drove the verification runs (ctest, `-DENABLE_MPI=OFF` build, clang-tidy) listed above.


evaleev:
Pushed a1b64dc6b on top of this branch, which simplifies `do_replicate`.

The quadratic traffic on master has a single cause: receivers `insert` into `local` while later ranks still have their turn to broadcast `begin()..end()`, so each root re-sends what it received from earlier roots. The minimal fix is to leave the loop as it was and just defer the inserts: receivers push `(key, value)` into a side buffer, and the buffer is drained into the container after the last rank's turn. Every rank then broadcasts exactly what it held on entry, and total traffic is N.

With that, the key snapshot + `find(acc, key)` re-lookup on the sender is unnecessary (it only existed because inserts were still happening mid-loop), so I dropped it. I also dropped the whole-shard `vector<pair<K,V>>` broadcast: on every receiver it holds the serialized byte buffer of the entire root shard and the deserialized vector alive at once, so peak transient memory is ~2x the largest shard on top of resident data — a problem for the 6-D GENTENSOR cases where a shard is multi-GB. Per-entry broadcast keeps the peak at essentially resident size (the side buffer is moved out, and `Tensor`-valued payloads are shallow handles anyway). Fewer collective round-trips is a separable optimization; if per-entry latency turns out to matter for a real workload it should be done as chunked batches with a byte budget, with a measurement, in its own PR.

The idempotency flag, its resets, and the `gop.min` collective decision are unchanged. `test_replication` no longer probes the archive layout for packed payloads (that only validated the bulk path); the `2 * nkeys` serialization bound alone distinguishes a linear replication from the old re-broadcasting one — I confirmed it fails with the master behavior reintroduced. `test_replication` at 1–4 ranks and `test_dc` at 1, 2, 4 ranks pass.

One thing to think about, not addressed here: the fast path silently skips propagation if a rank inserted locally after the last `replicate()`. The doc comment says "does not reconcile replicas", but nothing enforces it; a `gop.min`/`gop.max` on `size()` that falls back to a full replicate when they disagree would catch the common case for one extra small reduction.

evaleev (approved):

PR #816, opened 2026-09-18
mra: scale the truncation tolerance by the mean cell width, not the smallest

evaleev:
## Problem

`truncate_mode` 1, 2 and 3 scale the truncation tolerance by the physical width of the box, `tol*min(1, 2^-level * L)`. For an anisotropic cell the width has to be reduced to one number, and `L` was `cell_min_width`, the smallest dimension. That ties the tolerance to the cell's **aspect ratio** rather than to its resolution: at the level that resolves the widest dimension to a given spacing, the narrow dimension is already finer by the aspect ratio, and the tolerance is tightened by that factor for nothing.

Nobody noticed because every molecular MADNESS calculation goes through `set_cubic_cell`, where min and mean coincide. It bites slab and wire cells, which is what periodic MPQC uses.

## Evidence

MPQC/MRA, H10 chain, LDA/PBE/TPSS/HF, periodic cell 100x100x18 bohr, eps 1e-6, `truncate_mode` 0 -> 1:

| method | wall | peak RSS |
|---|---|---|
| LDA | 5m15s -> 1h11m | 6.3 -> 51.6 GB |
| PBE | 5m47s -> 1h14m | 6.3 -> 51.6 GB |
| TPSS | 7m15s -> 1h17m | 6.6 -> 51.9 GB |
| HF | 8m25s -> 1h29m | 6.3 -> 51.7 GB |

Hartree-Fock pays the same 10x as the functionals, so this is the tolerance, not the XC term. Running the same chain with *open* boundaries in the same cell separates the contributions: the narrow cell alone costs ~2x (1.4x in a cubic 200 bohr box vs 2.6x in this one), and periodicity multiplies that by ~4, because the operator apply screens lattice displacements with this very tolerance (`funcimpl.h`, `do_apply`), so images that were negligible survive.

## Fix

Use the geometric mean, `volume^(1/NDIM)` — the width consistent with the box volume. Cubic cells are unaffected.

Measured, MPQC LDA h10 chain, open BC, 400x400x40 cell, eps 1e-6, converged in 11 iterations either way:

| tolerance | energy | wall | peak RSS |
|---|---|---|---|
| mode 0 | -5.57595199 | 4m04s | 5.9 GB |
| mode 1, mean (this PR) | -5.57596852 | 6m13s | 9.6 GB |
| mode 1, min (before) | -5.57597161 | 9m59s | 14.4 GB |

1.6x less time *and* memory at aspect ratio 10, with the energy moving 3.1e-6 where mode 0 sits 2.0e-5 away from the old answer. I have not run a tighter reference, so I claim only that the two mode-1 values bracket a 3e-6 spread; the mean is not presented as *more* accurate than the min, just very much cheaper and far closer than mode 0. (The "before" numbers were taken with a temporary switch that is not part of this PR.)

`test_truncate_tol` covers modes 0-3 in a 400x400x40 cell and pins that a cubic cell is unchanged.


evaleev:
Fixed in 5c95a6980. `clear_cell()` now zeroes `cell_geometric_mean_width` alongside `cell_volume` and `cell_min_width`, so modes 1-3 cannot scale by a cell that is no longer set.

The finding was real: with the cached field left stale, `get_cell_geometric_mean_width()` returned 185.66 (the previous 400x400x40 cell) after `clear_cell()`, where `get_cell_min_width()` correctly returned 0. `test_truncate_tol` now covers the reset path and failed on that exact value before the fix.

One incidental note for anyone writing a similar test: the reset path has to be left with `set_cell()`, not `set_cubic_cell()`, since the latter writes into the existing cell tensor and `clear_cell()` has emptied it.

PR #818, opened 2026-09-20
mra: the rest-of-crystal operator, a lattice sum that excludes the home cell

evaleev:
## The rest-of-crystal operator

`OperatorInfo::images_only = true` builds a `SeparatedConvolution` that is the lattice sum with the home cell (L = 0) removed. Removing one lattice vector is not "drop R = 0 on every axis": that would also drop every image with a zero component, and for a chain periodic along z the only images at all are (0,0,n). The operator is instead a sum over the 2^p − 1 nonzero bit patterns of the p lattice-summed axes; in each pattern an axis whose bit is set sums R ≠ 0 and one whose bit is clear takes R = 0 only (`LatticeImages::{all, exclude_home, home_only}` on `Convolution1D`). Term by term the fit is the one the full lattice sum uses, so `full == home + images` holds exactly with the same fit and nothing cancels. Rank is (2^p − 1) × the fit's. `lattice_summed()` is unchanged, so the displacements stay on the non-periodic domain.

**Why.** MPQC forms the periodic Coulomb correction as V − V⁰: two full applies on the cuspy total charge density, whose difference is ~1e-11 for an isolated system, so neither input can be loosened (10× looser on one side moves the energy 3e-6). Built directly, the images potential is smooth and long-range, and the home-cell potential can then be split by density: the nuclear part is cached and only the smooth electronic density is applied per iteration.

Two things have to be handled for the identity to survive an infinite lattice sum and a non-cubic cell:

1. **Dropped low-exponent tail.** An infinite sum drops the fit's Gaussians with exponent < 0.25/L² (their lattice sum is a gauge constant). Their home-cell part is not flat, and a home operator built from the full fit keeps it. The images operator appends the dropped terms as home-only factors with negated coefficients, so it is exactly (lattice sum as computed) − (home cell with the full kernel). `make_coeff_for_operator` returns the tail alongside the kept fit. Without this the H10 chain (L = 18, 15 dropped terms) showed a uniform −3.7e-5 shift of every orbital energy.

2. **Shell-decay stop in `do_apply`.** The apply stops sweeping displacements at the first shell that contributes nothing, assuming a kernel that decays away from the source. The images kernel vanishes near the source and turns on near the lattice images; for a source 1.8 bohr from the periodic face of a 100×100×18 cell the empty short-axis shells ended the sweep before the wrapped displacement carrying the nearest-image interaction. The loss is exactly-empty blocks, hence independent of the threshold, of bmax and of the number of images, and it grows with the transverse size because long-axis shells sort later (identity residual 2.5e-3 in that cell vs 1e-7 in 18³; fit term 42 lost 4 % of its images potential). The kernel is translation-invariant but not lattice-periodic, so the wrapped displacement l − 2ⁿ is not equivalent to l as the lattice-modulated shells assume. `SeparatedConvolution::screen_by_shell_decay()` is false for `images_only`; `do_apply` then sweeps `get_disp_active(n)` — the displacements with a nonzero block at that level, built once per level and ordered by decreasing block norm (source-independent in the NS form) — and stops at the first displacement whose contribution is negligible for the source at hand. No assumption about how the kernel decays in space, exactly the blocks of a full sweep (identity residuals unchanged to the last digit), and cheap: on the 35 M-coefficient chain density the images apply is 1–3 s against 12–15 s for the home-cell operator. Other operators are untouched.

## Test

`test_images_operator` (short): `full == home + images` to 20·thresh for periodic-z N = 2 (1 pattern), periodic-xyz N = 1 (7 patterns), both with infinite range (kept rank 50 + 21 dropped → 71 and 371), plus the non-cubic near-face case (residual 3.1e-5 at thresh 1e-6, the displacement-reach floor that `full` and `home` share in that cell). `testsuite` passes.

## Validation in MPQC (z820, mode 1, eps 1e-6, Γ, LDA)

| | V − V⁰ | V^home + V^images |
|---|---|---|
| H10 isolated, 400³ cell: E | −5.5759481699316 | −5.5759481699342 |
| potential build / iter | 4.1 + 4.4 s | 0.56 + 0.22 s |
| H10 chain, period 18, 100×100×18: E | −5.6399995290 | −5.6399967565 |
| potential build / iter | 19.3 + 19.8 s | 1.81 + 1.42 s |
| wall, 20 iterations | 38:11 | 24:50 |

The remaining 2.8e-6 on the chain is at the level of the threshold-independent `bmax` reach floor of the apply in a 5.6:1 cell (free-space and lattice-summed potentials are both ~1e-4 off the analytic result there, cured by bmax 6–8), so at that level the two formulations disagree with each other rather than one with a reference.

https://claude.ai/code/session_01QSnAdQMbxZYShH9kLUjN1p


JonathonMisiewicz:
On my review list.

src/madness/mra/funcimpl.h
```diff
@@ -5123,6 +5123,9 @@ template<size_t NDIM>
             int nused = 1;  // Counts #used at each distance
             std::optional<double> real_last_distsq;
             std::optional<std::uint64_t> lattice_last_distsq;
+            // the shell-wise stop below assumes a kernel that decays away from the source; an operator
+            // whose kernel does not (the rest-of-crystal operator) opts out and sweeps every displacement
```
JonathonMisiewicz: I'd recommend specifying that rest-of-crystal operator _eventually_ decays monotonically, to make the use case clearly. We're not working with completely lawless operators.

src/madness/mra/funcimpl.h
```diff
@@ -5148,6 +5151,10 @@ template<size_t NDIM>
             for (const auto& displacement: displacements) {
               if (skip_predicate(displacement)) continue;
 
+              // without the shell-decay stop the list is ordered by decreasing block norm (get_disp_active),
```
JonathonMisiewicz: Could we possibly add a test that if we take an operator that normally does have shell stop and use this alternate ordering, we get the same result, but less efficiently?

JonathonMisiewicz:
I'm thrilled to see this PR! I had the rough idea a while ago but thought that the (2^p - 1) factor would make the computational expense prohibitive, so it wouldn't be worth the development effort.

Can you offer some brief comments on why we see the time reduction in the 1D case, and give some comparative timing benchmarks for the 2D case?

JonathonMisiewicz (commented):
Probably will have some more to add before long, figured I might as well post these now.

src/madness/mra/operatorinfo.h
```diff
@@ -68,6 +68,10 @@ struct OperatorInfo {
     std::vector<KernelRange> range = std::vector<KernelRange>(6);
     bool debug=false;
     std::optional<bool> truncate_lowexp_gaussians;  // if given, overrides the default for whether to truncate low-exponent gaussians
+    /// lattice-summed kernels only: sum over the images and exclude the home cell (L = 0),
```
JonathonMisiewicz: These are all relevant facts, but don't belong in this comment right above `images_only`. Simply specifying what parts of this comment apply when `images_only = true` would be very helpful.

src/madness/mra/convolution1d.h
```diff
@@ -255,6 +255,19 @@ namespace madness {
     /// interface for 1 term and for 1 dimension;
     /// the actual data are kept in ConvolutionData1D
     /// Derived classes must implement rnlp, issmall, natural_level
+    /// Which lattice translations a lattice-summed 1D kernel includes
```
JonathonMisiewicz:
```suggestion
    /// Derived classes must implement rnlp, issmall, natural_level,
    /// and which lattice translations a lattice-summed 1D kernel includes
```

src/madness/mra/operator.h
```diff
@@ -261,12 +279,33 @@ namespace madness {
             }
             // WARNING: discardG0 = true ignores the coefficients of truncated
             //          terms
+            // deep copies: Tensor assignment shares the buffer, and the truncation may edit
+            // coefficients in place (#820's truncate_mixed_expansion folds the tail into
```
JonathonMisiewicz:
```suggestion
            // coefficients in place (truncate_mixed_expansion folds the tail into
```
Mention of 820 is useful now, won't be useful later. Which reminds me - we should explicitly check that my folding doesn't cause a problem for this PR, if you haven't already.

src/madness/mra/funcimpl.h
evaleev: Done (2b651ca41): the comment now says the rest-of-crystal kernel decays monotonically only beyond the nearest image — it vanishes near the source — and points at get_disp_active() for what such an operator visits instead. Same wording on screen_by_shell_decay().

src/madness/mra/funcimpl.h
evaleev: Added (2b651ca41): SeparatedConvolution::set_screen_by_shell_decay(bool) overrides the default, and test_images_operator applies the free-space Coulomb operator both ways to the near-face density in the 100x100x18 cell: |difference| = 1.5e-14, 0.34 s through the norm-ordered visit vs 0.29 s with the shell stop. So: same result, only slower, as you predicted.

src/madness/mra/operatorinfo.h
evaleev: Cut to two lines (2b651ca41): what the flag does, and where the mechanics live (SeparatedConvolution::initialize_images_only).

src/madness/mra/convolution1d.h
evaleev: The real problem was placement: the enum and its doc had landed between Convolution1D's own doc comment and the class, so Doxygen and readers saw the two glued together. Moved the enum (with its doc) above Convolution1D's comment (2b651ca41); the "Derived classes must implement …" sentence is back to its own paragraph.

src/madness/mra/operator.h
evaleev: Reference dropped (2b651ca41). And yes, checked directly: the branch is now rebased on master with #820 in, so the two infinite-range identity cases run on top of truncate_mixed_expansion's folding — |full − home − images| = 7.3e-7 with full rank 61 and images 61 + 11 (the 11 are the terms by which the folded fit differs from the full one, on the full fit's exponents; the tail is computed as that difference, with deep copies since Tensor assignment shares the buffer and the folding edits in place). The rank check in the test became a range for the same reason.

evaleev:
Thanks — and the 2D question was the right one to ask: the first answer I had for it was not good, and fixing that is now in the PR.

**Why it is cheap.** Your intuition names the right terms with the roles swapped. The cost of a lattice-summed apply sits at the fine levels of the cuspy total density, and there only the *sharp* fit terms have non-empty blocks. For the images operator those terms' factors on a periodic axis exclude R = 0, so their only non-negligible blocks are at the wrapped displacements that reach a neighbouring image; every other block is empty (`issmall`) and the term is skipped for that node. The *diffuse* terms survive, but they act at the coarse levels, where there are few nodes. What is left is a coarse-level apply plus a thin skin at the periodic faces — and that holds for any number of periodic axes: with p axes the rank is (2^p − 1)× the fit's, but every pattern excludes the home cell along at least one axis, so the sharp terms vanish at fine levels in every pattern.

**What 2D exposed.** The norm-ordered visit is source-independent, and in 2D most of its large entries are wrapped displacements — blocks that matter for a source next to a face and land outside the cell for everyone else. A far-from-face node walked thousands of them per level and rejected each by validity, so the images apply was nearly as expensive as the lattice sum (73 s vs 85 s). `get_disp_active()` now groups the list by wrap pattern, each group carrying the box of sources that can reach at least one member, and `do_apply` skips the groups a source cannot reach (2b651ca41). Away from the faces only the unwrapped group is visited, which at fine levels is empty for this kernel.

**Numbers** (10-cusp neutral chain, eps 1e-6, mode 1, 8 threads; "cold" includes the one-time 1D operator setup at the wrapped displacements, "warm" is the per-apply cost an SCF pays each iteration):

| cell | p | ranks full / home / images | full | home | images cold / warm |
|---|---|---|---|---|---|
| 100×100×18, periodic z | 1 | 75 / 87 / 88 | 14.5 s | 14.7 s | 6.4 / **0.56 s** |
| 18×18×100, periodic xy | 2 | 74 / 86 / 235 | 69 s | 69 s | 2.3 / **2.9 s** |
| 18³, periodic xyz | 3 | 56 / 83 / 419 | 3.8 s | 3.2 s | 1.3 / **0.22 s** |
| 100×100×18, a cusp on the face | 1 | | 14.5 s | 14.4 s | 6.4 / **0.84 s** |

The last row is a case Ed raised: a nucleus on a facet is projected as two halves on the two faces, whose nearest-image interaction is at distance zero, so the images kernel is singular there and its short-range blocks are applied at every level near that face. It costs 50 % more than a chain inside the cell, not an order of magnitude, and the periodized chain's lattice-summed potential agrees with the free-space reference to 1.3e-4 (the isotropic displacement-reach floor in a 5.6:1 cell, which is #819's subject). A test for it is in (052ef4030).

evaleev:
@robertjharrison you probably want to glance at this also since this introduces yet another operator kernel ... but worth it since it greatly simplifies periodic case by splitting intra-cell interactions (which necessarily include self-interaction that requires careful handling, via "exchange" computation for smooth densities and mundane exclusion of self-pairing for point-like charges) from inter-cell interactions (small and rapidly-decaying except near the periodic boundaries).

PR #819, opened 2026-09-21
mra: make the reach of the displacement sweep isotropic in real space

evaleev:
## The reach of the displacement sweep is isotropic in real space

A level-n block of the nonstandard apply carries the kernel's roughness at the scale of the level-n box. `do_apply` visits displacements up to `bmax` boxes (4 in 3D) and leaves the rest to coarser levels, on the assumption that roughness decays fast with distance. In a non-cubic cell that assumption holds per axis only: a fit term that is rough along the long axis (width ≈ the long box) is smooth along a short axis and spans (long box)/(short box) boxes there, so its level-n block stays O(1) out to that many boxes — and no coarser level can supply it, because none can represent the long-axis roughness. The isotropic `bmax = 4` drops those blocks for good, a threshold-independent bias. (`bmax_default()` already said so: "the decision to have bmax be isotropic is only valid for hypercubes".)

Measured, free-space Coulomb potential of a neutral chain in a 100×100×18 cell at eps 1e-6: 1.3e-4 off the analytic result (1e-5 in the 18³ cell), unchanged at eps 1e-7, 1e-5 once the short axis reaches 4 long boxes.

`Displacements` now builds its lists with a per-axis reach `bmax_d = ⌈bmax_default · w_max / w_d⌉` — `bmax_default` on every axis of a cube, (4, 4, 23) in the cell above — and rebuilds them when `set_width()` changes the shape (a pure reorder when it does not). `make_disp` loses its hand-written per-dimension loops for an `IndexIterator` over per-axis extents; `make_disp_periodic` builds its wrapped translations per axis. The extra displacements are only ever norm-checked: for a decaying kernel the shell-wise stop ends the sweep at the same real distance as before. Cubic results are unchanged bit for bit; `testsuite` passes.

**Cost.** Lists 729 → 3807 (standard) and 1377 → 7533 (periodic, n = 6) in the 5.6:1 cell; the apply on a 35 M-coefficient density took 2 % longer, MPQC's per-iteration potential build on the H10 chain 6 %. Operators that cannot use the shell stop (the rest-of-crystal operator of #818) sweep their own norm-ordered list, so they are not affected by the list length.

**Test.** `test_anisotropic_reach` (short, 7 s): the Coulomb potential of five neutral blobs spanning 13 of the 18 bohr of the short axis against the analytic result — 1.2e-5 in 100×100×18 (5.8e-5 with the isotropic reach, which fails the 25·eps bound) and 5.2e-6 in 18³.

**Effect on applications.** MPQC syncs the displacement *ordering* to the cell (`math/mra/context.cpp`) but had no way to change the reach, so every non-cubic-cell MRA calculation so far carried this bias. On the H10 chain (period 18, 100×100×18, Γ, LDA, mode 1, eps 1e-6) the total energy moves by −1.04e-4 Ha, in both the V − V⁰ and the V^home + V^images formulation of the periodic potential; with the bias gone the two agree to 1.5e-7 Ha (they differed by 2.8e-6 before).

Independent of #818; either merge order is clean.

https://claude.ai/code/session_01QSnAdQMbxZYShH9kLUjN1p


JonathonMisiewicz:
It's intuitive to me that we want having the covered bmax be isotropic in real space. The comments, once again, baffle me:

I don't understand this language of "sweep" and "carrying the kernel's roughness." It's this idea of roughness that confuses me most. This idea of "a fit term that is rough along the long axis" confuses me. Our operators are all represented as linear combination of operators. True, some are singular, and those can only be approximated to finite precision, but those all have short-range roughness, not long-range (think 1/r). I don't know what long-range roughness of a single fit term it's talking about. How can a _gaussian_ be rough?!

evaleev:
Fair — both words were shorthand, and the comments now say the precise thing (9ecf232f1, description updated too).

**"Sweep"** = the loop in `FunctionImpl::do_apply` over the displacement list for one source node at level n: for each displacement l it applies the level-n non-standard block and accumulates into the node l boxes away. `bmax` bounds |l_d| in that list.

**"Rough"** was shorthand for *unresolved at level n on that axis*. In the non-standard form the operator acts at level n only through the blocks that involve a level-n wavelet — (d|s), (s|d), (d|d); the (s|s) block is the parent's. For a 1D Gaussian factor e^{−a x²} on boxes of width h those blocks scale like (a h²)^{k/2} while a h² ≪ 1 (wavelets are orthogonal to polynomials of degree < k) and are O(1) once a h² ≳ 1, i.e. once the Gaussian is narrow compared with the box. So a Gaussian is never rough in your sense; it is or isn't resolved at a given level, and in a non-cubic cell that is an axis-dependent statement because h differs per axis.

That is the whole mechanism: a term with a h_x² ≈ 1 on the long axis has a h_z² ≈ a h_x² (h_z/h_x)² ≪ 1 on the short one, so its (d_x|s_y|s_z) block is O(1) in x and decays along z only as the (s|s) matrix element, e^{−a (l_z h_z)²} — O(1) out to |l_z| ≈ h_x/h_z boxes, 5.6 in the 100×100×18 cell, beyond an isotropic `bmax = 4`. No other level contains that block, since a level-n wavelet is orthogonal to every coarser scaling function, so the loss is a threshold-independent bias (1.3e-4 in the potential there; 1e-5 with the per-axis bound). The relevant distance is the term's width 1/√a, isotropic in real space, hence `bmax_d ∝ 1/h_d`.

robertjharrison:
@evaleev @JonathonMisiewicz Note also the ability to apply derivatives of Gaussians with different (m=0,1,2) orders in different directions.  These kernels introduce anisotropy even in a cubic cell and also are non-monotonic until x>1/sqrt(2a) for the first derivative and x>sqrt(3/2a) for the second derivative.  Suggest you include testopdir in your testing if not already included.

evaleev:
> non-monotonic until x>1/sqrt(2a) for the first derivative and x>sqrt(3/2a) for the second derivative...

@robertjharrison nonmonotonicity is harder than the rest. How is nonmonotonic decay currently handled?

